PR #10, opened 2024-04-05
ci: update container from nextjs docs

conorsch:
Trying to get a more prod-compatible container image, rather than running the "dev" server as entrypoint.

New container spec was taken from [0]. Opts in to "standalone" output generation, and accordingly adds `sharp` to  the dependencies list, as specified in [1].

Also tacked on mention of `.envrc` because I use direnv locally to manage env vars.

[0] https://nextjs.org/docs/pages/building-your-application/deploying#docker-image
[1] https://nextjs.org/docs/pages/building-your-application/deploying#image-optimization

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+Containerfile
+.dockerignore
+node_modules
+npm-debug.log
+README.md
+.next
+.git
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ yarn-error.log*
 .env.development.local
 .env.test.local
 .env.production.local
+
+# Support direnv, as well
+.envrc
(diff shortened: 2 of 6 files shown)


src/utils/token/tokenFetch.tsx
```diff
@@ -47,6 +47,10 @@ export const fetchToken = async (
           }
         });
 
+        if (typeof tokenInner !== "string") {
+          tokenInner = uint8ArrayToBase64(tokenInner);
+        }
+
```
conorsch, line 53: This is an actual logic change, that warrants scrutiny. Makes sense to me based on my understanding of the code.

philipjames44 (approved):

conorsch:
@philipjames44 I'm not sure how disruptive this change would be to your existing workflow. In my local testing, it was straightforward to use this image to run a localhost:3000 server and use the app locally. If you have a strong preference for the old "dev" runtime, perhaps we could preserve that as `Containerfile.dev` or similar.

PR #20, opened 2024-05-24
wasm working

turbocrime:
cc @philipjames44 

this config works for me

proper package to be published soon, but you could use use https://github.com/penumbra-zone/web/pull/1158 if you want to build it locally


src/components/charts/buySellChart.tsx
turbocrime, line 1: note use of dynamic in this file. i couldn't get nextjs to be happy with the import any other way

turbocrime: recommend viewing with whitespace off, most of this diff is just indentation

turbocrime:
package has released https://www.npmjs.com/package/@penumbra-zone/wasm/v/5.0.0

PR #62, opened 2024-09-23
feat: header with connection button

### LINKED ISSUE #42: Implement ability to connect wallet

### LINKED ISSUE #45: UIv2 navigation

VanishMax:
Closes #42 

Closes #45 

Adds a header with navigation and connect/disconnect features. 

It is the copy from Minifront but without the sync bar and sync status popover. Do you think we need it here? If so, it is better to create a new issue for this since it would require setting up the state and requesting data.

Disconnected desktop: 

image

Connected desktop: 

image

Mobile (might need other icons for routes):

image


JasonMHasperhoven:
> It is the copy from Minifront but without the sync bar and sync status popover. Do you think we need it here? If so, it is better to create a new issue for this since it would require setting up the state and requesting data.

If so, let's proceed in any case and do it in steps.

JasonMHasperhoven (approved):

PR #68, opened 2024-09-24
Setup pages for v2 routes

JasonMHasperhoven:
As title says, this PR adds the pages for the v2 routes.

### DIFF
--- a/src/app/v2/page.tsx
+++ b/src/app/v2/page.tsx
@@ -1,17 +1,6 @@
-'use client';
-
-import { useEffect } from 'react';
-import { useRouter } from 'next/navigation';
+import { redirect } from 'next/navigation';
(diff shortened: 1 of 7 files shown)


src/app/v2/page.tsx
```diff
@@ -1,13 +1,17 @@
...
-    <section>
-      <Text h2>Hi!</Text>
-    </section>
-  )
-}
+const RedirectPage = () => {
+  const router = useRouter();
 
-export default HomePage;
+  useEffect(() => {
```
VanishMax, line 10: issue: this will only redirect when the page is loaded. Instead, use `redirect` function that will re-route on the server side: https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#redirect-function

VanishMax (changes requested):
one small but blocking issue

PR #70, opened 2024-09-25
feat: state management and sync popover

### LINKED ISSUE #66: Decide state manager for V2
Once you choose this, it will be forever... choose wisely.

_what can be unburdened by what has been_ - Tal

### LINKED ISSUE #67: Implement progress component
Reference: https://preview.ui.penumbra.zone/?path=/docs/ui-library-progress--docs

Will require https://github.com/penumbra-zone/dex-explorer/issues/66 to be implemented

VanishMax:
Closes #66 

Closes #67 

This PR sets up the Zustand as the state management library and showcases its usage with the sync feature


next.config.mjs
```diff
@@ -2,6 +2,14 @@
 
 const nextConfig = {
   transpilePackages: ["@penumbra-zone/protobuf"],
+  experimental: {
+    swcPlugins: [
+      [
+        "@preact-signals/safe-react/swc",
```
VanishMax, line 8: this plugin helps all react components understand if the signals are used inside it – and subscribe to signal updates

src/state/connection/index.ts
```diff
@@ -0,0 +1,62 @@
+import { signal } from '@preact-signals/safe-react';
+import { PenumbraRequestFailure, PenumbraState, PenumbraManifest } from '@penumbra-zone/client';
+import { penumbra, PRAX_ORIGIN } from '@/utils/penumbra';
+
+export const providerManifest = signal<PenumbraManifest>();
+
+export const providerConnected = signal<boolean>();
```
VanishMax, line 7: refactored the `useConnect` to use signals – this way `providerConnected` can be used not only within react components but also in any external TS code

src/state/status/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    syncPercent: status.catchingUp ? 0 : 1,
+    syncPercentStringified: status.catchingUp ? '0%' : '100%',
+  };
+};
+
+/**
+ * Receives the status stream from the view service
+ * and stores it in the `penumbraStatus` signal.
+ */
+export const streamPenumbraStatus = async () => {
```
VanishMax, line 43: refactored status streaming to, in my opinion, more readable state. Also learned a lot how the streaming works

grod220:
Curious your thoughts on [Dan Abramov's comment](https://github.com/facebook/react/issues/26704#issuecomment-1522044060). Is this implementation of signals getting us into a place that will hurt future upgrades?

[Mobx](https://mobx.js.org/README.html#introduction) has a quite similar, signal-esc API. Should we consider that instead for less future risk? 

Also, I wonder what state management libraries will be compatible with server components if/when we use this.

VanishMax:
@grod220 You might be right about Signals – it might truly be broken in the future, my oversight. I still believe it's React's problem that it cannot support this API without hacks.

Anyway, after a couple more thoughts, I tend towards Zustand instead of Mobx. They both are great in terms of the component updates only when necessary but Mobx discourages me with its `observer` HOC and the need to pass the state instance as props – usually this means having global state.

Before changing this PR, I propose we use Zustand but differently from Penumbra Minifront. We should not try to implement global state with Zustand but keep it localized by feature. Let's say we have a feature "sync status" that requests data and stores it in the state. It would be best to place UI, api and state in the same directory, and keep it as independent as possible. Having the global state makes it mostly uncontrollable and more difficult to test. Local by-feature state would help decouple the code and make it more scalable and future proof.

As for the server components, idk if they even need state management. So far I've only seen the minimal request-and-render server components.

PR #56, opened 2024-09-19
feat: support PENUMBRA_INDEXER_CA_CERT env var override

conorsch:
Adds opt-in support for declaring a CA certificate value for the database TLS connection. Some managed postgres solutions, such as DigitalOcean's, require this setting. In order to use it:

  1. set the PENUMBRA_INDEXER_CA_CERT env var with the string contents of the db's CA
  2. remove the `sslmode=require` from the connection auth string

The need for 2 is because the `connectionString` param clobbers any manual `ssl` opts in the db config, and is documented here [0].

Closes #55.

[0] https://node-postgres.com/features/ssl#usage-with-connectionstring

PR #71, opened 2024-09-25
docs: add quickstart docs for dev env

conorsch:
Adds a nix config so that tooling like `pnpm` is available immediately. Fleshes out the documentation around connecting to managed database instances, since merge of #56.


JasonMHasperhoven (approved):
Nice

PR #74, opened 2024-09-26
Add path value support for env PENUMBRA_INDEXER_CA_CERT

JasonMHasperhoven:
Add support to pass a path name as value for the env PENUMBRA_INDEXER_CA_CERT

### DIFF
--- a/src/utils/indexer/connector.tsx
+++ b/src/utils/indexer/connector.tsx
@@ -1,8 +1,11 @@
 // @ts-nocheck
 /* eslint-disable -- disabling this file as this was created before our strict rules */
-import { Pool } from "pg";
-import { BlockInfo, LiquidityPositionEvent } from "./types/lps";
-import { bech32ToInner } from "../math/bech32";
+import { Pool } from 'pg';
+import fs from 'fs';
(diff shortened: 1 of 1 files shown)


src/utils/indexer/connector.tsx
```diff
@@ -14,11 +17,13 @@ export class IndexerQuerier {
...
+      ...(ca && {
         ssl: {
-            rejectUnauthorized: true,
-            ca: process.env.PENUMBRA_INDEXER_CA_CERT,
-          },
+          rejectUnauthorized: true,
+          ca: ca.startsWith('-----BEGIN CERTIFICATE-----')
+            ? ca
+            : fs.readFileSync(process.env.PENUMBRA_INDEXER_CA_CERT, 'utf-8'),
+        },
```
conorsch, line 26: This is a bit automagicky for my tastes: I'd prefer discrete, explicitly named env vars for filepath or content of the cert. But I appreciate that you were careful to preserve the pre-existing behavior, so if this works for you, I'm happy to merge it!

conorsch (approved):

PR #75, opened 2024-09-26
Chart

JasonMHasperhoven:
Implements the lightweight-charts lib of tradingview.

- Moved the candles fetching logic with the merging to a new `candles` api route. The implementation is slightly updated to start at the end height and go backwards.

image


src/app/v2/trade/chart/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Token } from '@/utils/types/token';
+import { useCandles } from '@/fetchers/candles';
+
+const { colors } = tailwindConfig.theme.extend;
+
+interface ChartProps {
+  height: number;
+}
+
+const asset1: Token = {
```
JasonMHasperhoven, line 13: Temporarily hardcoded

src/pages/api/blocks/[...heights].ts
```diff
@@ -0,0 +1,26 @@
...
+const indexerEndpoint = process.env['PENUMBRA_INDEXER_ENDPOINT'] ?? '';
+if (!indexerEndpoint) {
+  throw new Error('PENUMBRA_INDEXER_ENDPOINT is not set');
+}
+
+interface QueryParams {
+  heights?: string[];
+}
+
+export default async function blockInfoFetchHandler(req: NextApiRequest, res: NextApiResponse) {
```
Comment (VanishMax) [inline src/pages/api/blocks/[...heights].ts:13]: suggestion: move this into the `app` directory while applying the newer NextJS's API.

I believe we should discontinue supporting the `pages` directory and move everything related to v2 to the `app`. It's not just the new API but also, if done correctly, a performance increase.

src/app/v2/trade/chart/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  inner: '',
+};
+const asset2: Token = {
+  display: 'GM Wagmi',
+  symbol: 'gm',
+  decimals: 5,
+  inner: '',
+};
+
+export function Chart({ height }: ChartProps) {
```
VanishMax, line 26: suggestion/nitpick (non-blocking for now): move the `chart` directory into the `components`. Even though it is tempting to store all components related to the trade page nearby, I think we should separate it

package.json
```diff
@@ -54,8 +55,10 @@
     "framer-motion": "^11.3.31",
     "grpc-tools": "^1.12.4",
     "grpc_tools_node_protoc_ts": "^5.3.3",
+    "lightweight-charts": "^4.2.0",
     "lodash": "^4.17.21",
     "lucide-react": "^0.445.0",
+    "milliseconds": "^1.0.3",
```
VanishMax, line 61: question: is this library used in the PR? What's the benefit of it?

src/fetchers/candles.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { CandlestickData } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
+
+export const useCandles = (
+  symbol1: string,
+  symbol2: string,
+  startBlock: number | undefined,
+  limit: number,
+) => {
+  return useQuery({
+    queryKey: ['candles', startBlock],
```
VanishMax, line 11: question: is this key array enough to cache correctly?

Asking cause I'm just learning react-query and truly want to understand its caching mechanisms. The [docs mention](https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#array-keys-with-variables) a serializable object with key parameters like `useQuery({ queryKey: ['todo', 5, { preview: true }], ...})` that help identifying the correct query. Should it be smt like this?

```suggestion
    queryKey: ['candles', { symbol1, symbol2, startBlock, limit }],
```

Also, @grod220 asked about the server caching before. According to [NextJS docs](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching#fetching-data-on-the-server-with-the-fetch-api), using the `fetch` API should result in correct caching by the framework. Maybe we don't even need to set up the cache by react-query since it would only take the user memory space?

VanishMax (approved):
All good!

Issue #73, opened 2024-09-25
Asset registry not correct for testnet-phobos-2


conorsch:
There's a new testnet chain `penumbra-testnet-phobos-2`, running here: https://testnet.plinfra.net There's also a dex-explorer instance running against that chain https://dex-explorer.testnet.plinfra.net (#72). Despite recent updates to the prax registry in https://github.com/prax-wallet/registry/pull/91 & https://github.com/prax-wallet/registry/pull/93, as well as corresponding changes in this repo to consume https://github.com/penumbra-zone/dex-explorer/issues/57, the `test_usd` asset still isn't loading correctly:

![dex-explorer-undefined-asset](https://github.com/user-attachments/assets/75675769-73ae-468a-9de7-507d322de899)

The asset on the lefthand side is `test_usd`, or at least that's how it was specified on when swapped via `pcli`. But the app fails to understand the asset, which indicates that we likely have the wrong string set in the prax registry still: https://github.com/prax-wallet/registry/blob/7cfdd0000c4f4c458e147b85ba6c7f3f87066c11/input/chains/penumbra-testnet-phobos-2.json#L144 Let's debug that and fix it.

conorsch:
With an assist from @hdevalence, I've confirmed that the `test_usd` asset string appears to be correct. One can cross-check it by using the `AssetMatadataById` RPC: https://buf.build/studio/penumbra-zone/penumbra/penumbra.core.component.shielded_pool.v1.QueryService/AssetMetadataById?target=https%3A%2F%2Ftestnet.plinfra.net&selectedProtocol=grpc-web So why isn't it displaying correctly in the application? Are other assets displaying correctly? At least `UM` is. Let's test with more.

conorsch:
Ah-ha, it looks like a problem with the container build story. Testing locally via `pnpm dev` with the same database backend, I see the assets rendering correctly:

![dex-explorer-dev-env-working](https://github.com/user-attachments/assets/f4033244-2bb2-4720-9ecd-5c78932bb59c)

So something with the `npm run build` is emitting bad or incomplete artifacts.

conorsch:
[...]

PR #76, opened 2024-09-26
build: use dev runtime for container image

conorsch:
As a temporary measure, uses the `pnpm run dev` environment inside the container image. This is an anti-pattern. We *should* be using the "standalone" output of nextjs and running that directly via `node`, but doing so does not permit runtime configuration of the chain-id, used for loading assets, which is must-have for the built image.

Sprinkles in a few `justfile` additions to make testing locally a bit more straightforward, as we'll surely be exercising these config paths again in the near future.

Refs #73.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,18 @@
-Containerfile
-.dockerignore
-node_modules
-npm-debug.log
-README.md
-.next
-.git
+# ignore everything by default
+**
+
+# allow JS files
+!*.ts
+!*.js
+!*.mjs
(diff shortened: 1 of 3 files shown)


JasonMHasperhoven:
Here we should revert to using the build command, but we should be able to provide different envs to the docker container and create multiple images, each for a different environment.

conorsch:
Unanimously we do not want to use a dev build in the live deployments; see #80 for best alternative to satisfy #73. Closing this out as unwanted.

PR #77, opened 2024-09-27
Mobx spike

grod220:
Closes https://github.com/penumbra-zone/dex-explorer/issues/66
Closes https://github.com/penumbra-zone/dex-explorer/issues/67

Takes over from: https://github.com/penumbra-zone/dex-explorer/pull/70

An experiment of using Mobx for the connection + streaming status state. Observations:

Pros
- No need for selectors/useShallow anymore. Just import what you need. Optimizations built in.
- Classes where you can mutate variables feel like the better syntax versus an immutable object w/ function fields
- Feels like less boilerplate (?). Interfaces do not need to be defined and class types are inferred.
- Effects built in w/ `when()` and don't need to enter react context to form store relationships + reactions

Cons
- Remembering to wrap components with HOC `observer()` if relying upon observable vars
- Actions often must be indicated. Sometimes can be inferred with `makeAutoObservable()`, but often have to make explicit setter functions. Adding decorator syntax could help with this.


src/state/connection/index.ts
```diff
@@ -1,52 +1,60 @@
...
 
-export const useConnectionState = create<ConnectionState>()((set, get) => ({
-  connected: false,
-  manifest: undefined,
+  constructor() {
+    makeAutoObservable(this);
+
+    if (typeof window !== 'undefined') {
+      this.setup();
+    }
```
grod220, line 14: A substitute for on mount useEffect() hook. Downside is it needs a window guard to ensure it only runs in the client.

src/state/status/index.ts
```diff
@@ -1,82 +1,78 @@
...
-    latestKnownBlockHeight: status.catchingUp ? undefined : status.fullSyncHeight,
-    syncPercent: status.catchingUp ? 0 : 1,
-    syncPercentStringified: status.catchingUp ? '0%' : '100%',
-  };
-};
+    when(
+      () => connectionStore.connected,
+      () => void this.setup(),
+    );
+  }
```
grod220, line 36: A substitute for https://github.com/penumbra-zone/dex-explorer/pull/77/files#diff-4c50d9dc437e2361505ebbcb744f036646280707449adeff19752b15f236c992L28-L33. Able to define store-to-store relationships/reactions outside of react.

src/components/header/provider-popover.tsx
```diff
@@ -1,19 +1,13 @@
...
-
-const connectionSelector = (state: ConnectionState) => ({
-  manifest: state.manifest,
-  disconnect: state.disconnect,
-});
+import { connectionStore } from '@/state/connection';
 
 export const ProviderPopover = () => {
-  const { manifest, disconnect } = useConnectionState(useShallow(connectionSelector));
+  const { manifest, disconnect } = connectionStore;
```
grod220, line 10: Ah, an example of where I forgot to wrap a component in Observer. This would result in this not being reactive to those observable vars.

VanishMax (approved):
praise: it is a great example of Mobx! I appreciate that you had experience with it and showed the use cases far better than what I could find in the skimming of the docs.

I don't oppose the Mobx now. Let's see what the rest of the team has to say

src/components/header/provider-popover.tsx
VanishMax: so this should probably be fixed before merging?

src/state/connection/index.ts
```diff
@@ -1,52 +1,60 @@
...
+  }
 
-  connect: async () => {
+  async connect() {
     try {
       await penumbra.connect();
     } catch (error) {
       if (error instanceof Error && error.cause) {
         if (error.cause === PenumbraRequestFailure.Denied) {
           // TODO: replace these alerts with toasts
```
TalDerei, line 45: could we create a separate issue to track this?

src/components/header/provider-popover.tsx
TalDerei:
```
export const ProviderPopover = observer(() => {
    ...
}
```

I didn’t find any other instances of missing observer.

src/state/status/index.ts
```diff
@@ -1,82 +1,78 @@
...
+
+  private setUnaryStatus(status: StatusResponse) {
+    this.loading = false;
+    this.error = undefined;
+    this.syncing = status.catchingUp;
+    this.fullSyncHeight = status.fullSyncHeight;
+    this.latestKnownBlockHeight = status.catchingUp ? undefined : status.fullSyncHeight;
+    this.syncPercent = status.catchingUp ? 0 : 1;
+    this.syncPercentStringified = status.catchingUp ? '0%' : '100%';
```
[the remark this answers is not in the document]
TalDerei: we should keep these [react rendering optimizations](https://mobx.js.org/react-optimizations.html) in mind as a reference moving forward.

TalDerei (approved):

TalDerei:
from my perspective, using Mobx for managing client-side app state seems to be exactly what we're looking for in a HFT DEX UI since it focuses on fine-grained reactivity, keeping the UI reactive with a performant batch update system while intelligently avoiding unnecessary re-renders.

Issue #72, opened 2024-09-25
Add automatic testnet deployments


conorsch:
We're now running a version of the DEX explorer for the testnet chain `penumbra-testnet-phobos-2`, hosted here: https://dex-explorer.testnet.plinfra.net/. Let's re-enable continuous deployment of that environment on merge into main.

PR #78, opened 2024-09-27
Remove the use of a raw indexer

### LINKED ISSUE #49: Remove dependency on raw events database
We want to remove the dependency of this code on the raw events database, instead using a derived database created with `pindexer`.

cronokirby:
Closes #49.

Apologies for the big PR.

This removes the use of the old indexer, replacing it instead with the new indexing database, derived after running pindexer.

This also introduces some amount of typing into the API calls, using Zod Schemas.
I decided to not put also playing around with TRPC or some other typed API system here, to avoid having too much in a single pull request.

I also updated the frontend as necessary to use the new typed api routes I introduced; as a consequence I rewrote the component displaying the events of a position, as a consequence removing a lot of code.


conorsch:
This is huge. I defer to web team on the suitability of the refactoring, but lgtm. Making this change will require some comms because a new db backend is required. In order to evaluate these changes on the testnet, we must resolve #73, so I'd prefer that change lands first, so we have working asset display on the old db backend on the testnet instance, then we can proceed with updating the deployment to use the new db and confirm no regressions or breakage. 

Regarding release comms, I think we can be a lot more explicit about the db backend in the README, but we don't need to solve that problem now.

conorsch:
> In order to evaluate these changes on the testnet, we must resolve https://github.com/penumbra-zone/dex-explorer/issues/73, so I'd prefer that change lands first

We have a candidate PR for resolving #73 in #80 now, which is undergoing review.

conorsch:
We have automatic deployments to testnet again (#72). Switching to a different schema on the backend will be a breaking change, but it looks like this PR won't be landing soon, given the conflicts. Just holler at me when we're closer to the finish line, and I'll make sure the backend databases are configured correctly on the testnet-facing deployment.

PR #79, opened 2024-10-01
Dynamic client side envs

JasonMHasperhoven:
As per @conorsch 's request, this adds support to set envs dynamically post-build.

- Routes need to pass `envs` from getServerSideProps as props
- _app will take this and provide it as a context (in v2 we could use mobx or whatever state manager we decide on)
- Components can consume useEnvContext to access the envs


JasonMHasperhoven:
I'm still on the fence whether we should do this, @grod220 any thoughts?

src/pages/explorer/index.tsx
```diff
@@ -33,4 +40,14 @@ export default function Explorer() {
       </Box>
     </Layout>
   );
-}
\ No newline at end of file
+}
+
+export async function getServerSideProps() {
+  const envs = getClientSideEnvs();
```
VanishMax, line 46: suggestion: instead of calling this function on every page, expose selected environmental variables with `NEXT_PUBLIC_` prefix. See [nextjs docs about runtime envs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#runtime-environment-variables).

src/utils/env/getClientSideEnvs.ts
```diff
@@ -0,0 +1,13 @@
+export function getClientSideEnvs() {
+  const whitelist: string[] = ['PENUMBRA_CHAIN_ID', 'PENUMBRA_CUILOA_URL'];
```
VanishMax, line 2: suggestion: add `.env.example` file to env. It shouldn't have any values, just keys with comments on what they do, and where to get them from

src/utils/env/getClientSideEnvs.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  const envs = whitelist.reduce(
+    (envs: Record<string, string>, key) => ({
+      ...envs,
+      [key]: process.env[key] ?? '',
```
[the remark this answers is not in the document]
JasonMHasperhoven: We need to store it somewhere since we receive it from the server-side, this is a temporary solution as we're still debating the state manager.

src/pages/explorer/index.tsx
JasonMHasperhoven: NEXT_PUBLIC_ envs will be compiled within the build unfortunately. So you can not provide an env var starting with NEXT_PUBLIC_  after the build is complete and let nextjs use that.

grod220:
It's tough for me to review this with all the formatting changes. Could you open another PR that just does the formatting (and adds CI/CD checks for prettier)?

JasonMHasperhoven:
> It's tough for me to review this with all the formatting changes. Could you open another PR that just does the formatting (and adds CI/CD checks for prettier)?

I didn't wanna do it, but I did it: https://github.com/penumbra-zone/dex-explorer/pull/80

Closing this one.

JasonMHasperhoven:
Closing this one.

grod220:
> I didn't wanna do it, but I did it: https://github.com/penumbra-zone/dex-explorer/pull/80

This isn't the formatting change PR though right? I think those changes are important and there should be a CI/CD test like in penumbra-zone/web.

grod220:
If we merged that separately, this PR would have automatically been cleared up

PR #80, opened 2024-10-02
Dynamic client side envs 2

JasonMHasperhoven:
As per @conorsch 's request, this adds support to set envs dynamically post-build.

- ~~Routes need to pass envs from getServerSideProps as props~~
- ~~_app will take this and provide it as a context (in v2 we could use mobx or whatever state manager we decide on)~~
- ~~Components can consume useEnvContext to access the envs~~

---

- Client now makes a request to /api/env to get the clientSideEnv
- Refactored app to use fetchers (similar to minifront's design) -- which consume useRegistry & useEnv


src/pages/_app.tsx
```diff
@@ -31,9 +32,11 @@ function app({ Component, pageProps }: AppProps) {
   });
   return (
     <ChakraProvider theme={theme}>
-      <Analytics />
-      <SpeedInsights />
-      <Component {...pageProps} />
+      <EnvContext.Provider value={pageProps.envs}>
```
JasonMHasperhoven, line 35: Meant as a placeholder until we have finalized a state manager.

conorsch:
Deploying from this tag, without updating the env vars, I see:

```
⨯ Error: Registry not found for 
     at Ct.get (/app/node_modules/.pnpm/@penumbra-labs+registry@11.3.1/node_modules/@penumbra-labs/registry/dist/index.js:2:113241) 
     at h (/app/.next/server/chunks/215.js:8:6274) 
     at d (/app/.next/server/chunks/215.js:8:6293) 
     at c (/app/.next/server/chunks/215.js:8:6607)
     at b (/app/.next/server/pages/trade/[[...params]].js:1:2857)
 ```
 
Which indicates that we no longer have sane default values. Can we add a default value for `PENUMBRA_CHAIN_ID="penumbra-1"`? Or at least a more informative error message that indicates a required environment variable should have been set and wasn't. Remember that RC is running a deployment, as well, and changes like the lack of defaults will cause that instance to fail when it's upgraded, unless we preserve defaults.
 
After setting the env vars properly, according to the new env vars, I confirmed that the assets were loading correctly for the testnet deployment, which is magnificent.

**Review by conorsch — CHANGES_REQUESTED** [review]
Works with the new env vars set! Asking for a bit more error-handling around the new vars: let's provide defaults, or at least log which required env vars weren't set if they're missing. See example error message in comments above.

File: src/components/pairSelector.tsx [hunk 2]
```diff
@@ -15,13 +15,13 @@ export default function PairSelector({
...
+  const tokenAssetsList = useTokenAssets();
   const [tokenAssets, setTokenAssets] = useState<Record<string, Token>>({});
   const [selectedAssets, setSelectedAssets] = useState<Token[]>([]);
 
   useEffect(() => {
-    const tokenAssets = fetchAllTokenAssets();
-    setTokenAssets(Object.fromEntries(tokenAssets.map(asset => [asset.symbol, asset])));
-  }, []);
+    setTokenAssets(Object.fromEntries(tokenAssetsList.map(asset => [asset.symbol, asset])));
+  }, [tokenAssetsList]);
```
grod220, line 24: This should be in react-query

src/pages/block/[block_height].tsx
```diff
@@ -704,3 +706,13 @@ export default function Block() {
...
+
+export async function getServerSideProps() {
+  const envs = getClientSideEnvs();
+
+  return {
+    props: {
+      envs,
+    },
+  };
+}
```
Comment (grod220) [inline src/pages/block/[block_height].tsx:718]: Two thoughts:

- Can we put this at a kind of parent layout component so we don't have to do this for every page we create (and need to remember or else risk errors)?
- What if we created an backend API route for returning this and we put it in react-query (client side) instead?

src/utils/token/tokenFetch.tsx
```diff
@@ -35,31 +34,40 @@ export const fetchAllTokenAssets = (): Token[] => {
   return tokens;
 };
 
+export function useTokenAssets() {
+  const envs = useEnvContext();
+  return fetchAllTokenAssets(envs['PENUMBRA_CHAIN_ID'] ?? '');
+}
```
grod220, line 40: As this is, I'm pretty sure that fetchAllTokenAssets() does not cache. If env context was in react-query, we could compose within react-query to handle caching globally for this work. Otherwise, every component's use of useTokenAssets() will be a new instance and re-running on every render.

src/pages/_app.tsx
grod220: We we anticipate this env context vars to be critical to a number of things in client side code, we should use react-query to fetch and sync its state with a mobx `EnvContextStore`.

grod220 (commented):
There is a lot in this pull request that opens a number of questions on how we want to handle data fetching within this app:
- client vs server requests
- async vs sync requests
- global caching & performance
- react-query syncing w/ state manager
- using a state manager versus react-context

Think it is worthwhile for others on the web team to take a look and we discuss in today's meeting

src/components/pairSelector.tsx
JasonMHasperhoven: Correct, but let's keep that to another PR (also this is v1 code).

JasonMHasperhoven:
Changing the implementation to use a request to fetch the client-side env.

JasonMHasperhoven:
It's ready for review again. I indeed like this implementation better @grod220 & added defaults for the client side @conorsch.

grod220 (approved):
Talked offline:
- For hooks consumed by deprecated code, label as deprecated. Given they don't consume loading/error states, can keep simple as needed.
- For hooks consumed by new code, should follow data/loading/error propagation pattern
- No nulls (if possible)

PR #85, opened 2024-10-03
feat: #65: pair selector

### LINKED ISSUE #65: Asset Selector
Designs are not currently firm for this.

Screenshot 2024-09-23 at 6 45 31 PM

Some things to consider:

- The current [Uiv2 AssetSelector](https://preview.ui.penumbra.zone/?path=/docs/ui-library-assetselector--docs) component is flexible. We should be able to re-purpose for this.
- Will need to be able to select asset in + asset out

Inspiration:

Dolomite:
Screenshot 2024-09-23 at 6 46 44 PM

DyDx
Screenshot 2024-09-23 at 6 47 04 PM

VanishMax:
Closes #65 

Implements `PairSelector` – a components that allows selecting the in- and out-assets. Features:
- Asset list is prefilled on the server using the Prax registry
- On provider connection, the stores fetch account assets and balances
- Fetching happens using `react-query` to cache the results
- All data is stored in corresponding stores, the `pair` store keeps the selected assets for the future use by other components

image

image


src/state/shared/assets.ts
```diff
@@ -0,0 +1,36 @@
...
+      () => void this.setup(),
+    );
+  }
+
+  async setup() {
+    try {
+      runInAction(() => this.loading = true);
+      const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
+      this.assets = responses.map(getDenomMetadata);
+      this.loading = false;
```
grod220, line 29: issue: Async mobx actions are a bit tricky. It looks like you are doing `runInAction()` only for one of the fields. My suggestion is to create a `setLoading()` and `setAssets()` and `setError()`. Otherwise, you'll get a warning from mobx in the console for async functions.

src/state/shared/assets.ts
```diff
@@ -0,0 +1,36 @@
...
+  /** The error is set in case of a request failure */
+  error?: string;
+  assets: Metadata[] = [];
+
+  constructor() {
+    makeAutoObservable(this);
+
+    when(
+      () => connectionStore.connected,
+      () => void this.setup(),
```
grod220, line 20: question: I did this pattern on the connection status/block sync indicator PR. As it is on every page, it can always run. However, if we continue this pattern with all of our stores, won't that make a connection action request possibly too much when the view they are looking at doesn't require it?

Aka, should we instead have a useEffect() hook within react that initiates the setup? So that the lifecycle of the component governs the data loading?

VanishMax: The way I see it, the stores are initiated right where they are imported for the first time. This way, if defined and used correctly (we can define the rules about it), the stores will set up when needed or triggered by the user

src/state/shared/assets.ts
VanishMax: Realized the loading state was unnecessary. As for the rest, applied the setters as you proposed

VanishMax:
@grod220 do you mind to review it again?

src/shared/state/assets.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Metadata } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import { getDenomMetadata } from '@penumbra-zone/getters/assets-response';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { Constants } from '@/shared/configConstants';
+import { penumbra } from '@/shared/penumbra';
+import { connectionStore } from '@/shared/state/connection';
+import { queryClient } from '@/shared/queryClient';
+
+const fetchAssets = () => {
+  return queryClient.fetchQuery({
```
VanishMax, line 12: used react-query to apply caching for the fetched data but not using `useQuery` since there are multiple requests in this store – couldn't figure out how to merge them together

src/shared/state/pair.ts
```diff
@@ -0,0 +1,21 @@
+import { makeAutoObservable } from 'mobx';
+import { AssetSelectorValue } from '@penumbra-zone/ui/AssetSelector';
+
+class PairState {
+  from?: AssetSelectorValue;
+  to?: AssetSelectorValue;
```
VanishMax, line 6: the idea of this store is to be used by other components like charts and order book

src/shared/state/assets.ts
grod220, line 1: Think we still have an issue where server fetching should be handled by react-query for both assets and balances store. It would look something like:

```ts
// QueryClientProvider wraps whole app

export const PairSelector = observer(
    const { data } = useAssets();
...
}

import { connectionStore } from '@/shared/state/connection';
export const useAssets = () => {
  return useQuery({
    queryKey: ['assets'],
    queryFn: async () => {
      const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
      return responses.map(getDenomMetadata);
    },
    enabled: connectionStore.connected,
  });
};
```

VanishMax: Thanks for the feedback! I refactored it to use only react-query instead of mobx to keep the state of assets and balances. It feels like a good solution

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+    queryKey: ['assets'],
+    enabled: connectionStore.connected,
+    queryFn: async () => {
+      const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
+      return responses.map(getDenomMetadata);
+    },
+  });
+
+  return connectionStore.connected ? accountAssets : registryAssets;
+};
```
grod220, line 32: Suggestion:

```ts
export const useAssets = () => {
  const { data: registry, error: registryErr } = useRegistry();

  const registryAssets = useQuery({
    queryKey: ['registry-assets'],
    enabled: Boolean(registry),
    queryFn: () => {
      if (!registry) {
        throw new Error('Registry not available');
      }
      return registry.getAllAssets();
    },
  });

  const accountAssets = useQuery({
    queryKey: ['view-service-assets'],
    enabled: connectionStore.connected,
    queryFn: async () => {
      const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
      return responses.map(getDenomMetadata);
    },
  });

  return connectionStore.connected
    ? accountAssets
    : { ...registryAssets, error: registryAssets.error ?? registryErr };
};
```

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
+import { ViewService } from '@penumbra-zone/protobuf';
+import { getDenomMetadata } from '@penumbra-zone/getters/assets-response';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { penumbra } from '@/shared/penumbra';
+import { connectionStore } from '@/shared/state/connection';
+import { useQuery } from '@tanstack/react-query';
+import { useEnv } from '@/fetchers/env';
+
+export const useAssets = () => {
```
grod220, line 9: issue: we should have a code comment above to let folks know about the connected vs non-connected nature of the asset fetcher.

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { data: env, isLoading } = useEnv();
+  const chainId = !isLoading ? env?.PENUMBRA_CHAIN_ID : process.env['PENUMBRA_CHAIN_ID'];
+
+  const registryAssets = useQuery({
+    queryKey: ['registry-assets'],
+    queryFn: () => {
+      const registryClient = new ChainRegistryClient();
+      const registry = registryClient.bundled.get(chainId ?? '');
```
grod220, line 17: issue: think we should default to using the remote registry unless we know we are using something that never changes (like stakingAssetId)

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getDenomMetadata } from '@penumbra-zone/getters/assets-response';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { penumbra } from '@/shared/penumbra';
+import { connectionStore } from '@/shared/state/connection';
+import { useQuery } from '@tanstack/react-query';
+import { useEnv } from '@/fetchers/env';
+
+export const useAssets = () => {
+  const { data: env, isLoading } = useEnv();
+  const chainId = !isLoading ? env?.PENUMBRA_CHAIN_ID : process.env['PENUMBRA_CHAIN_ID'];
```
grod220, line 11: thought: I didn't know the client has both client-side env variables and remote ones we fetch (that overlap). This feels a bit brittle and feels like we should really only have one or the other.

Regardless, I think the logic of choosing which env vars are correct should be in `useEnv` and not have to be dealt with by other hooks (and add a code comment about the two different env vars there).

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+    queryKey: ['registry-assets'],
+    queryFn: () => {
+      const registryClient = new ChainRegistryClient();
+      const registry = registryClient.bundled.get(chainId ?? '');
+      return registry.getAllAssets();
+    },
+  });
+
+  const accountAssets = useQuery({
+    queryKey: ['assets'],
```
grod220, line 23: suggestion: would update this query key to `view-service-assets`

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+    enabled: connectionStore.connected,
```
grod220, line 24: docs-related: because this hook relies upon an observable, all consumers of this hook must be wrapped in an `observer()` component. Let's add that to the code comment above.

src/shared/state/assets.ts
grod220, line 17: issue: let's throw instead of defaulting to a `""` for chain id. Aka:

```ts
if (!chainId) {
  throw new Error("No chain id to retrieve registry")
}
const registry = registryClient.bundled.get(chainId);
```

That means you'd likely also need to add an `enabled: Boolean(chainId)` field.

src/shared/state/assets.ts
```diff
@@ -0,0 +1,32 @@
...
+    queryFn: async () => {
+      const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
+      return responses.map(getDenomMetadata);
+    },
+  });
+
+  return connectionStore.connected ? accountAssets : registryAssets;
```
grod220, line 31: issue: we should be propagating errors also from the querys this hook consumes. That means we have to combine the states. If you look into `useRegistry()` there is an example of this. Would look like:

```ts
  return connectionStore.connected
    ? accountAssets
    : { ...registryAssets, error: registryAssets.error ?? registryErr };
```

src/shared/state/balances.ts
```diff
@@ -0,0 +1,17 @@
...
+import { BalancesResponse } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+import { penumbra } from '@/shared/penumbra';
+import { connectionStore } from '@/shared/state/connection';
+import { useQuery } from '@tanstack/react-query';
+
+const fetchQuery = async (): Promise<BalancesResponse[]> => {
+  return Array.fromAsync(penumbra.service(ViewService).balances({}));
+};
+
+export const useBalances = () => {
```
grod220, line 11: docs: we should had a code comment that specifies that consumers of this hook are required to be wrapped in an `observer()` func

grod220 (approved):
Lots of comments about the react-query hooks, but after addressing feel free to merge. Realizing we are kinda discovering best practice patterns for composing react-queries together.

src/shared/state/assets.ts
grod220: thought: I think, we don't need to propagate the loading state as if `enabled: false` I think `registryAssets.isLoading` is true (?). Not sure should check that. If the loading state isn't propagated via the enabled field of registryAssets, we should need to do the same as: https://github.com/penumbra-zone/dex-explorer/blob/df94a11f1d0d8e8fcdddeaa238471da719676696/src/fetchers/registry.ts#L29. 

However, registryAssets is in a loading state while `enabled: false`, we can do as I suggested above and simplify the pattern like my example in useRegistry()

VanishMax:
Applied all PR comments, documented the result. Merging 🚀

PR #90, opened 2024-10-09
ci: add automatic deploys to testnet environment

### LINKED ISSUE #73: Asset registry not correct for testnet-phobos-2

conorsch:
Now that we've resolved #73, we're unblocked from restoring rolling deployments on merge to main, to the testnet instance [0]. This new workflow blocks on completion of the build-container logic, then bounces the pre-existing deployment to pull the most recently built container.

Refs #72.

[0] https://dex-explorer.testnet.plinfra.net

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,55 @@
+---
+# Workflow for deploying the DEX explorer webapp https://github.com/penumbra-zone/dex-explorer
+# Bounces a container deployment, to repull the latest image.
+name: deploy dex-explorer
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-container:
+    name: build container
+    uses: ./.github/workflows/container.yml
+    secrets: inherit
+    permissions:
+      contents: read
+      packages: write
+
+  deploy-testnet:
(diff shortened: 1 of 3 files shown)


conorsch:
CI is passing. I cannot run this workflow until it exists in main, so I'm going to self-merge and inspect the run results when it gets triggered, and follow up as necessary.

PR #91, opened 2024-10-09
ci: only build container once per merge

conorsch:
Since the deploy workflow depends on the container workflow, we'll trigger the latter on running the former. Therefore the container workflow shouldn't run on its on when main changes.

Also cleans up the post-deploy wait-for task to use proper shell commands.

Follow-up to https://github.com/penumbra-zone/dex-explorer/pull/90. Refs #72.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,10 +1,12 @@
 ---
 name: Build container image
 on:
-  # By default, build on merge to main.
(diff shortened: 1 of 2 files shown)


conorsch:
Works, you can see a successful deploy here: https://github.com/penumbra-zone/dex-explorer/actions/runs/11262943204

Issue #64, opened 2024-09-23
Route Book


grod220:
Use existing API route

<img width="434" alt="Screenshot 2024-09-23 at 6 36 25 PM" src="https://github.com/user-attachments/assets/e372b9be-9e7d-4bd3-85c4-fd56e1af542a">

cronokirby:
I think this API route will have to be related to simulations, and not to what positions exist, if you want indirect route information and to aggregate by price.

PR #93, opened 2024-10-11
Route Book

JasonMHasperhoven:
Implements https://github.com/penumbra-zone/dex-explorer/issues/64

With very basic UI (and without mid price)
image


src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+const querier = new DexQueryServiceClient({
+  grpcEndpoint: process.env['PENUMBRA_GRPC_ENDPOINT'] ?? '',
+});
+
+interface Params {
+  symbol1: string;
+  symbol2: string;
+  hops: string;
+  limit: string;
+}
```
Comment (grod220) [inline src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts:17]: idea: not for this PR, but think trpc would be good for us to use

src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  const regex = new RegExp(`^${symbol}$`, 'i');
+  return metadata.find(asset => regex.test(asset.symbol));
+}
+
+export async function GET(_request: Request, context: { params: Params }) {
+  const { symbol1, symbol2, hops: hopsParam, limit: limitParam } = context.params;
+  const hops = Number(hopsParam);
+  const limit = Number(limitParam);
+
+  const registry = await chainRegistryClient.remote.get(process.env['PENUMBRA_CHAIN_ID'] ?? '');
```
Comment (grod220) [inline src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts:29]: suggestion: if env vars are not correctly injected, think this will produce a strange error. Suggest:

```ts
  const chainId = process.env['PENUMBRA_CHAIN_ID'];
  if (!chainId) {
    throw new Error('No chain id in env vars');
  }
  const registry = await chainRegistryClient.remote.get(chainId);
```

src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  const metadata: Metadata[] = registry.getAllAssets();
+
+  const metadata1 = getMetadataBySymbol(metadata, symbol1);
+  const metadata2 = getMetadataBySymbol(metadata, symbol2);
+
+  const sellSidePair = new DirectedTradingPair({
+    start: metadata2?.penumbraAssetId,
+    end: metadata1?.penumbraAssetId,
+  });
```
Comment (grod220) [inline src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts:38]: suggestion: we should probably type guard for this too right (to prevent strange error downstream)?

```ts
  if (!metadata1?.penumbraAssetId || !metadata2?.penumbraAssetId) {
    throw new Error(`no asset id for ${symbol1} or ${symbol2}`);
  }
```

src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+    end: metadata2?.penumbraAssetId,
+  });
+
+  const data = await Promise.all([
+    querier.liquidityPositionsByPrice(sellSidePair, hops),
+    querier.liquidityPositionsByPrice(buySidePair, hops),
+  ]).then(([asks, bids]) => ({
+    asks: asks?.slice(0, limit),
+    bids: bids?.slice(0, limit),
+  }));
```
Comment (grod220) [inline src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts:50]: suggestion: I find it more readable if we don't mix async/await & then()

```ts
const [asks, bids] = await Promise.all([
  querier.liquidityPositionsByPrice(sellSidePair, hops),
  querier.liquidityPositionsByPrice(buySidePair, hops),
]);

const data = {
  asks: asks?.slice(0, limit),
  bids: bids?.slice(0, limit),
};
```

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+interface RouteWithTotal extends Route {
+  total: number;
+}
+
+function getTotals(data: Route[], isBuySide: boolean, limit: number): RouteWithTotal[] {
+  const totals: RouteWithTotal[] = Object.values(
+    data.reduce((displayData: Record<string, RouteWithTotal>, route: Route) => {
+      const key = route.price;
+      return {
+        ...displayData,
```
grod220, line 33: thought: I wonder if this spread, creating a new object on every loop, can be inefficient on large data sets. Maybe not a concern. If so though, can do something like `const displayData = new Map<number, RouteWithTotal>();` with a normal for loop.

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  position: Position;
+  price: number;
+  amount: number;
+}
+
+interface RouteWithTotal extends Route {
+  total: number;
+}
+
+function getTotals(data: Route[], isBuySide: boolean, limit: number): RouteWithTotal[] {
```
grod220, line 28: thought: wonder if it would help to have a code comment above about what this is doing

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function getDisplayData(
+  data: Position[],
+  computePositionId: ((position: Position) => PositionId) | undefined,
+  asset1: Metadata | undefined,
+  asset2: Metadata | undefined,
+  isBuySide: boolean,
+  limit: number,
+): RouteWithTotal[] {
```
grod220, line 60: thought: wonder if making this an object instead of ordered args would improve the readability of the use of this func

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  if (!computePositionId || !asset1 || !asset2) {
+    return [];
+  }
+
+  const routes = data
+    .filter(position => position.state?.state.toLocaleString() === 'POSITION_STATE_ENUM_OPENED')
```
grod220, line 66: issue:

```diff
- .filter(position => position.state?.state.toLocaleString() === 'POSITION_STATE_ENUM_OPENED')
+ .filter(position => position.state?.state === PositionState_PositionStateEnum.OPENED)
```

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+    .map(position => {
+      const direction =
+        asset1.penumbraAssetId?.inner &&
+        position.phi?.pair?.asset2?.inner &&
+        (position.phi.pair.asset2.inner as unknown as string) ===
+          uint8ArrayToBase64(asset1.penumbraAssetId.inner)
```
grod220, line 72: issue: we should use the equality functions offered within the bufbuild types and almost never touch the inner. Aka:

```ts
position.phi.pair.asset2.equals(asset1.penumbraAssetId)
```

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+          ? -1
+          : 1;
+
+      const asset1Exponent = getDisplayDenomExponent(asset1);
+      const asset2Exponent = getDisplayDenomExponent(asset2);
+
+      const getValue = (property: Amount | undefined, exponent: number) =>
+        fromBaseUnit(BigInt(property?.lo ?? 0), BigInt(property?.hi ?? 0), exponent);
```
grod220, line 80: issue: This feels like it should be defined outside the loop

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  asset1: Metadata | undefined,
+  asset2: Metadata | undefined,
+  isBuySide: boolean,
+  limit: number,
+): RouteWithTotal[] {
+  if (!computePositionId || !asset1 || !asset2) {
+    return [];
+  }
+
+  const routes = data
```
grod220, line 65: question: where did all the domain knowledge for the routes come from? Did it come from the core repo? Can we add documentation on these steps or code comments to help the next person?

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+      return {
+        lpId: bech32Id,
+        position,
+        price,
+        amount,
+      };
+    })
+    .filter(displayData => displayData.amount > 0)
+    .sort((a, b) => b.price - a.price) as Route[];
```
grod220, line 109: thought: should we use `toSorted()`?

src/fetchers/book.ts
```diff
@@ -0,0 +1,28 @@
...
+  asks: Position[];
+  bids: Position[];
+}
+
+export const useBook = (
+  symbol1: string | undefined,
+  symbol2: string | undefined,
+  hops: number | undefined,
+  limit: number | undefined,
+) => {
```
grod220, line 14: idea: Not critical to this PR, but starting to realize we'll want to include block height in the query key with nearly all our queries (as they'll get invalidated every block and refetched). Maybe something for later.

src/fetchers/book.ts
```diff
@@ -0,0 +1,28 @@
...
+    queryKey: ['book', symbol1, symbol2, hops, limit],
+    queryFn: async (): Promise<BookResponse> => {
+      if (!symbol1 || !symbol2 || !limit) {
+        return {
+          asks: [],
+          bids: [],
+        };
+      }
+      const res = await fetch(`/api/book/${symbol1}/${symbol2}/${hops}/${limit}`);
+      return (await res.json()) as BookResponse;
```
grod220, line 25: thought: This is where trpc would be helpful

src/shared/round.ts
```diff
@@ -0,0 +1,17 @@
+import lodashRound from 'lodash/round';
+
+/**
+ * Replacement for toFixed()
+ *
+ * toFixed falsely rounds the number:
+ * > 1.0005.toFixed(3)
+ * 1.000 // instead of 1.001
+ */
+export function round(value: number, n: number, toFixed = true): number | string {
```
grod220, line 10: suggestion: we probably don't need toFixed option no?

src/shared/useComputePositionId.ts
```diff
@@ -0,0 +1,27 @@
...
+  PositionId,
+} from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
+
+export function useComputePositionId() {
+  const ref = useRef<(position: Position) => PositionId>();
+  const [, setIsSet] = useState(false);
+
+  useEffect(() => {
+    const set = async () => {
+      // cant import directly without breaking the build cmd
```
grod220, line 13: question: why does that happen I wonder. Is it because the wasm file is made for web and not nodejs?

src/shared/useComputePositionId.ts
```diff
@@ -0,0 +1,27 @@
...
+      const { computePositionId } = await import('@penumbra-zone/wasm/dex');
```
grod220, line 14: suggestion: given this is an sync thing, can we move this into react-query?

src/shared/round.ts
grod220: thought: should we write a few tests for this?

grod220 (commented):
Lots of comments, but feel free to merge after addressing

src/shared/round.ts
JasonMHasperhoven: We need it to display a consistent decimal length in the route book, otherwise some rows might display a shorter amount of numbers and make it all misaligned.

JasonMHasperhoven: Too simple for tests? Seems unnecessary imo

src/shared/useComputePositionId.ts
JasonMHasperhoven: I don't know the reason tbh..

src/components/route-book/index.tsx
JasonMHasperhoven: This doesn't work since the enum is a 1:
image

src/shared/useComputePositionId.ts
JasonMHasperhoven: Done

src/components/route-book/index.tsx
JasonMHasperhoven: Done

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  const asset1Exponent = getDisplayDenomExponent(asset1);
+  const asset2Exponent = getDisplayDenomExponent(asset2);
+
+  const getValue = (property: Amount | undefined, exponent: number) =>
+    fromBaseUnit(BigInt(property?.lo ?? 0), BigInt(property?.hi ?? 0), exponent);
+
+  const routes = data
+    // .filter(position => position.state?.state.toLocaleString() === 'POSITION_STATE_ENUM_OPENED')
```
grod220, line 84: nit: remove

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    .filter(position => position.state?.state === PositionState_PositionStateEnum.OPENED)
+    .map(position => {
+      const direction = position.phi?.pair?.asset2?.equals(asset1.penumbraAssetId) ? -1 : 1;
```
grod220, line 87: question: does `-1` and `1` have any inherent meaning within trade routing? If not, think it would be easier to make this a string label like `primary` and `secondary` or something.

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+}
+
+export function RouteBook() {
+  const { data: assets } = useAssets();
+  const asset1 = assets?.find(asset => asset.symbol === 'UM');
+  const asset2 = assets?.find(asset => asset.symbol === 'GM');
+  const asset1Exponent = asset1 ? getDisplayDenomExponent(asset1) : 0;
+  const asset2Exponent = asset2 ? getDisplayDenomExponent(asset2) : 0;
+  const { data: computePositionId } = useComputePositionId();
+  const { data } = useBook(asset1?.symbol, asset2?.symbol, 100, 50);
```
grod220, line 128: suggestion: related to my other comment, but I think we should consider:

```ts
if (!asset1?.symbol || !asset2?.symbol) {
   return <div>No asset selected</div>
}
```

Note that this comes before `useBook()`, and given the rule of hooks, you'd have to create another component after the if clause.

src/fetchers/book.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface BookResponseJson {
+  asks: JsonValue[];
+  bids: JsonValue[];
+}
+
+export const useBook = (
+  symbol1: string | undefined,
+  symbol2: string | undefined,
```
grod220, line 17: suggestion: think these should be required fields

src/fetchers/book.ts
```diff
@@ -0,0 +1,39 @@
...
+  hops: number | undefined,
+  limit: number | undefined,
+) => {
+  return useQuery({
+    queryKey: ['book', symbol1, symbol2, hops, limit],
+    queryFn: async (): Promise<BookResponse> => {
+      if (!symbol1 || !symbol2 || !limit) {
```
grod220, line 24: question: limit is a required field too? If so, think we should change the type signature of the hook.

grod220 (approved):
Some final suggestions, but feel free to merge after

src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts
Comment (VanishMax) [reply src/app/api/book/[symbol1]/[symbol2]/[hops]/[limit]/route.ts:12]: @grod220 @JasonMHasperhoven wouldn't it be easier to not use the `api` directory but a simple server-side fetch within a component? Like

```tsx
export default async function Page() {
  let data = await fetch('https://api.vercel.app/blog')
  let posts = await data.json()
  return (
    <ul>
      {posts.map((post) => (
        <li key={post.id}>{post.title}</li>
      ))}
    </ul>
  )
}
``` 

Instead of the `fetch` from my example, we could write `useAsksBids()` from the API made by Jason. It would do all the server-side work while sharing the types eliminating the need in TRPC. Wdyt?

src/components/route-book/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    .filter(displayData => displayData.amount > 0)
+    .toSorted((a, b) => b.price - a.price) as Route[];
+
+  return getTotals(routes, isBuySide, limit);
+}
+
+export function RouteBook() {
+  const { data: assets } = useAssets();
+  const asset1 = assets?.find(asset => asset.symbol === 'UM');
+  const asset2 = assets?.find(asset => asset.symbol === 'GM');
```
VanishMax, line 124: suggestion: use `pair` MobX state (`src/shared/state/pair.ts`) to get the asset info. You may also fill this state initially, so `from` and `to` won't be `undefined`

JasonMHasperhoven: This doesn't seem to work for me, the asset selector doesn't seem to fire on click?

src/components/route-book/index.tsx
JasonMHasperhoven: I see your point, if you don't mind I will address this in the next update (separate pr).

src/components/route-book/index.tsx
VanishMax: That's weird, it works fine for me... Would love if you can make the reproduction

https://github.com/user-attachments/assets/49067ddb-1ed5-4fce-8542-d2b6fb273e6b

Issue #95, opened 2024-10-15
Invalidate react-query cache upon new block


grod220:
We want our data sources to be refetched whenever a block advances. Todo:

1. Consider refactoring status indicator store to export blockHeight in a way that initiates on load and not only onConnect of Prax
2. Add blockHeight (from Mobx) to react-query fn key for:
  - Candle query
  - Routebook query

PR #99, opened 2024-10-24
Refetch on new block

grod220:
Closes https://github.com/penumbra-zone/dex-explorer/issues/95

Introduces new hooks that subscribe to the latest block and triggers refetching of queries whenever a new block is produced.


src/app/v2/layout.tsx
```diff
@@ -2,13 +2,12 @@
...
 import { ReactNode } from 'react';
 import { PenumbraUIProvider } from '@penumbra-zone/ui/PenumbraUIProvider';
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { QueryClientProvider } from '@tanstack/react-query';
 import { Display } from '@penumbra-zone/ui/Display';
 import { Header } from '@/components/header';
 import { SyncBar } from '@/components/header/sync-bar';
 import { enableStaticRendering } from 'mobx-react-lite';
-
-const queryClient = new QueryClient();
```
grod220, line 11: Removing the redundant query clients around the app

src/shared/state/compactBlock.ts
grod220, line 1: A whole lot of hooks are required to pull this off:
- useRpcChoices() to get the registry RPCs (if they are not connected to Prax)
- useGrpcTransport() to get the right transport (extension versus grpc web depending on connection status)
- useWalletOnWrongChain() to ensure they are connected to the site's matching chain
- useLatestBlockHeight() which retrieves the latest block + useStream() to ensure cache is always updated
- and finally useRefetchOnNewBlock() as a the utility that other hooks can use

src/pages/trade/ui/chart.tsx
```diff
@@ -21,86 +22,88 @@ const ChartLoadingState = ({ height }: ChartProps) => {
...
-  baseAsset,
-  quoteAsset,
-}: {
-  height: number;
-  baseAsset: Metadata;
-  quoteAsset: Metadata;
-}) => {
-  const chartElRef = useRef<HTMLInputElement>(null);
-  const chartRef = useRef<IChartApi | null>(null);
+const ChartData = observer(
```
grod220, line 25: Using hook that observes mobx connection state store. Hennce the wraps for two components with `observer`.

src/shared/api/env/env.ts
```diff
@@ -1,13 +1,6 @@
...
 
-export const useEnv = () => {
-  return useQuery({
-    queryKey: ['clientEnv'],
-    queryFn: async (): Promise<ClientEnv> => {
-      const res = await fetch('/api/env');
-      return (await res.json()) as ClientEnv;
-    },
-    staleTime: Infinity,
-  });
```
grod220, line 12: Moving away a bit from composing too much with hooks. QueryFn's that depend on env can just do an async await for it. While it may seem like multiple fetches will occur and we don't take advantage of react-query caching, it actually gets cached just fine with http caching in-browser.

src/shared/state/compactBlock.ts
```diff
@@ -0,0 +1,97 @@
+'use client';
```
VanishMax, line 1: issue (conventions): use kebab-case in the file naming

src/shared/useStream.ts
```diff
@@ -0,0 +1,55 @@
+'use client';
```
VanishMax, line 1: issue (conventions): use kebab-case in the file naming

src/shared/state/compactBlock.ts
VanishMax: suggestion: move the files from `state` to `shared/api`. Feels like the right choice for methods requesting data

VanishMax (approved):
Looks very neat! Don't want to block your work, so approving with comments regarding file structure

PR #113, opened 2024-10-25
refactor: apply FSD code structure

VanishMax:
This PR is only 1 line because I accidentally pushed 2 commits into `main` and cannot force-push their removal:
1. https://github.com/penumbra-zone/dex-explorer/commit/e13b25826389ccc5f708777ec5f59b3c77f1331e
2. https://github.com/penumbra-zone/dex-explorer/commit/6486f4c8460e3ddde13397386fff62b917404cb1

Let's use this PR for discussion of the proposed structure and refactoring, and for changes.

What is done:
1. `app` is moved to the root, `pages` directory in the root is created to distract NextJS from `src/pages`
4. all page and api code is moved to `src` and simply exported from the `app`
5. re-organized some files within other directories, renamed some files to kebab-cases
6. added documentation on the code structure

Main ideas:

- To comply with the NextJS file structure, the `app` and `pages` directories are used in the root of the project. However, it should only export the code defined in other directories.
- The `src` directory mainly follows the [Feature-Sliced Design (FSD)](https://feature-sliced.design/docs/get-started/overview) methodology. It defines the layers, slices and segments instead of splitting the files by its type (e.g. "components" or "state").
- All file names should use **kebab-case**.
- When using the FSD layers, apply the top-down approach: start by using the `pages` layer and extract the code from it only when this code is being shared. For example, the project initially may only be one `home` slice in the `pages` layer. When we add the `about` page, the `header` component defined in `home` might be moved 1 layer down to `widgets` or even to `shared` layer.
- Each slice might but not required to include these slices: `ui` for storing components, styles and images, `api` for interactions with the backends or extensions, `model` for storing functions related to business logic or data storage.
- Ideally, the slices must not import other slices from the same layer. In practice, we might allow doing so on the `entities` layer.
- Each slice must include the `index.ts` file representing the public API. To avoid the tree shaking issues, this file should define named exports (e.g. `export { A } from './ui/a.tsx` is good but `export * from './a.tsx` is not).


JasonMHasperhoven (approved):
LGTM, would it be possible to just create empty folders for features & entities with example folders (of features/entities) that are applicable to the dex (with something that we will build in the future)?

TalDerei:
LGTM

VanishMax:
@JasonMHasperhoven empty folders do not show up in git but I'll find another way to showcase the features and entities in the next PRs

Issue #94, opened 2024-10-15
Route book hops


grod220:
Display hops within route book. Design pending, can be scrappy.

Look for `ExecutionTraces` in the pb type

<img width="158" alt="Screenshot 2024-10-15 at 6 19 57 PM" src="https://github.com/user-attachments/assets/236642c8-2d80-4ea9-b194-6676bc3ab2f4">

PR #117, opened 2024-10-30
Route book -> SimulateTradeRequest refactor

grod220:
Closes https://github.com/penumbra-zone/dex-explorer/issues/64, https://github.com/penumbra-zone/dex-explorer/issues/94
A follow up from https://github.com/penumbra-zone/dex-explorer/pull/93

Previously was querying liquidity positions, but as we need to get the hops, converted the implementation to `SimulateTradeRequest`. Also moved a lot of parsing logic to the backend. 

Various boy-scouting refactors, sorry about that.


src/pages/trade/ui/chart.tsx
```diff
@@ -71,7 +71,7 @@ const ChartData = observer(
     }, [chartElRef]);
 
     useEffect(() => {
-      if (chartRef.current && !isLoading) {
+      if (chartRef.current && !isLoading && candles) {
```
grod220, line 74: On error, the casting on https://github.com/penumbra-zone/dex-explorer/pull/117/files#diff-c44d16aa7a593e9bbbb5d025963ae793cd30f7c22f703d4b5d4e195e766f20ebR83 causes a bug. Adding the guard here to prevent it.

src/pages/trade/ui/route-book.tsx
```diff
@@ -126,70 +11,56 @@ const RouteBookLoadingState = () => {
...
+              <td className='text-right tabular-nums'>{trace.total}</td>
+              <td className='text-right tabular-nums'>{trace.hops.length}</td>
+            </tr>
+          ))}
+          {multiHops.buy.map(trace => (
+            <tr key={trace.price + trace.total} style={{ color: 'green' }}>
+              <td className='text-left tabular-nums'>{trace.price}</td>
+              <td className='text-right tabular-nums'>{trace.amount}</td>
+              <td className='text-right tabular-nums'>{trace.total}</td>
+              <td className='text-right tabular-nums'>{trace.hops.length}</td>
```
grod220, line 40: Hops now visible. Final designs can be implemented in follow up PR.

Screenshot 2024-10-30 at 2 55 36 PM

src/shared/api/indexer/connector.tsx
```diff
@@ -1,11 +1,11 @@
-// @ts-nocheck
-/* eslint-disable -- disabling this file as this was created before our strict rules */
```
grod220, line 2: The lack of types here was pretty brutal as it wasn't detecting the import changes needed. Went ahead and added types.

src/shared/api/server/token-fetch.tsx
```diff
@@ -8,7 +8,8 @@ const getRegistry = (chainId: string): Registry => {
   return registryClient.bundled.get(chainId);
 };
 
-export const fetchAllTokenAssets = (chainId: string): Token[] => {
+// TODO: Deprecated. Use remote ChainRegistryClient directly.
+export const fetchAllTokenAssets_deprecated = (chainId: string): Token[] => {
```
grod220, line 12: Want to ensure no one uses this as it creates its own `Token` type primitive and uses bundled registry vs remote registry client.

src/shared/utils/base64.ts
grod220, line 1: A whole lot of utilities available in the npm packages seemed to have been manually copied over. Deleting them.

src/pages/trade/ui/chart.tsx
JasonMHasperhoven: 👍

src/shared/api/server/book/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        sell: sellMulti.filter(t => t.hops.length === 2),
+      },
+      multiHops: { buy: buyMulti, sell: sellMulti },
+    }),
+  );
+}
+
+const simulateTrade = async (querier: SimulationQuerier, req: SimulateTradeRequest) => {
+  try {
+    return await querier.simulateTrade(req);
```
JasonMHasperhoven, line 96: Just curious, why do you do try/catch instead of .catch() here?

JasonMHasperhoven (approved):
Nicely done 👍

src/shared/api/server/book/index.ts
grod220: If we are doing a sequence of .then().catch(), that makes sense. However, with async/await syntax it is more idiomatic to wrap in try catch instead of mixing styles.

PR #124, opened 2024-11-04
feat: mobile view of route book

### LINKED ISSUE #103: Mobile view of Route book
Screenshot 2024-10-24 at 4 30 37 PM

vacekj:
fixes #103

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "react": "^18.3.1",
     "react-chartjs-2": "^5.2.0",
     "react-dom": "^18.3.1",
+    "react-loading-skeleton": "^3.5.0",
     "react-outside-click-handler": "^1.3.0",
     "sharp": "^0.33.5",
     "webpack": "^5.94.0",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "react-dom": "^18.3.1",
     "react-outside-click-handler": "^1.3.0",
     "sharp": "^0.33.5",
-    "usehooks-ts": "^3.1.0",
     "webpack": "^5.94.0",
     "zod": "^3.23.8"
   },
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "react": "^18.3.1",
     "react-chartjs-2": "^5.2.0",
     "react-dom": "^18.3.1",
-    "react-loading-skeleton": "^3.5.0",
     "react-outside-click-handler": "^1.3.0",
     "sharp": "^0.33.5",
     "webpack": "^5.94.0",
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "react-dom": "^18.3.1",
     "react-outside-click-handler": "^1.3.0",
     "sharp": "^0.33.5",
+    "usehooks-ts": "^3.1.0",
     "webpack": "^5.94.0",
     "zod": "^3.23.8"
   },
--- a/src/pages/trade/model/trace.ts
+++ b/src/pages/trade/model/trace.ts
@@ -1,6 +1,5 @@
 import { Trace } from '@/shared/api/server/book/types.ts';
 
-/* TODO: write tests for this*/
 export const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
   if (!sellOrders.length || !buyOrders.length) {
     return;
--- a/src/pages/trade/model/trace.ts
+++ b/src/pages/trade/model/trace.ts
@@ -1,5 +1,6 @@
 import { Trace } from '@/shared/api/server/book/types.ts';
 
+/* TODO: write tests for this*/
 export const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
   if (!sellOrders.length || !buyOrders.length) {
     return;
--- a/src/pages/trade/model/trace.ts
+++ b/src/pages/trade/model/trace.ts
@@ -1,4 +1,5 @@
 import { Trace } from '@/shared/api/server/book/types.ts';
+import { round } from '@/shared/utils/numbers/round.ts';
 
 export const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
   if (!sellOrders.length || !buyOrders.length) {
@@ -20,8 +21,8 @@ export const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
   const spreadPercentage = (spread / midPrice) * 100;
 
   return {
-    amount: spread.toFixed(8),
-    percentage: spreadPercentage.toFixed(2),
-    midPrice: midPrice.toFixed(8),
+    amount: round(spread, 8),
+    percentage: round(spreadPercentage, 2),
+    midPrice: round(midPrice, 8),
   };
 };
--- a/src/pages/trade/model/trace.ts
+++ b/src/pages/trade/model/trace.ts
@@ -1,0 +1,27 @@
+import { Trace } from '@/shared/api/server/book/types.ts';
+
+export const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
+  if (!sellOrders.length || !buyOrders.length) {
+    return;
+  }
+
+  const lowestSell = sellOrders[sellOrders.length - 1];
+  const highestBuy = buyOrders[0];
+
+  if (lowestSell === undefined || highestBuy === undefined) {
+    return;
+  }
+
+  const sellPrice = parseFloat(lowestSell.price);
+  const buyPrice = parseFloat(highestBuy.price);
+
+  const spread = sellPrice - buyPrice;
+  const midPrice = (sellPrice + buyPrice) / 2;
+  const spreadPercentage = (spread / midPrice) * 100;
+
+  return {
+    amount: spread.toFixed(8),
+    percentage: spreadPercentage.toFixed(2),
+    midPrice: midPrice.toFixed(8),
+  };
+};
--- a/src/pages/trade/ui/route-book.tsx
+++ b/src/pages/trade/ui/route-book.tsx
@@ -83,22 +83,37 @@ export const ROUTEBOOK_TABS = [
   { label: 'Route Depth', value: 'depth' },
 ];
 
-const SkeletonRow = () => (
-  <tr className={`group relative h-[33px] border-b border-border-faded`}>
-    <td>
-      <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
-    </td>
-    <td className='relative text-xs text-right text-white'>
-      <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
-    </td>
-    <td className='relative text-xs text-right text-white'>
-      <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
-    </td>
-    <td className='relative text-xs text-right'>
-      <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
-    </td>
-  </tr>
-);
+const SkeletonRow = (props: { isSpread: boolean }) =>
+  props.isSpread ? (
+    <tr>
+      <td colSpan={4} className='border-y border-border-base'>
+        <div className='flex items-center justify-center gap-2 px-3 py-3 text-xs'>
+          <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+
+          <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+
+          <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+
+          <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+        </div>
+      </td>
+    </tr>
+  ) : (
+    <tr className={`group relative h-[33px] border-b border-border-faded`}>
+      <td>
+        <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+      </td>
+      <td className='relative text-xs text-right text-white'>
+        <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+      </td>
+      <td className='relative text-xs text-right text-white'>
+        <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+      </td>
+      <td className='relative text-xs text-right'>
+        <div className='w-full h-[22px] bg-neutral-800 rounded animate-pulse ml-auto'></div>
+      </td>
+    </tr>
+  );
 
 const RouteBookData = observer(({ bookData }: { bookData?: RouteBookResponse }) => {
   const multiHops = bookData?.multiHops;
@@ -154,9 +169,9 @@ const RouteBookData = observer(({ bookData }: { bookData?: RouteBookResponse }) 
                   ))}
                 </>
               ) : (
-                Array(16)
+                Array(17)
                   .fill(1)
-                  .map((_, i) => <SkeletonRow key={i} />)
+                  .map((_, i) => <SkeletonRow isSpread={i === 8} key={i} />)
               )}
             </tbody>
           </table>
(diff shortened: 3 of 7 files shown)


src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
   return (
-    <div>
-      <div className='text-gray-500'>Loading...</div>
+    <div className='flex items-center gap-1 py-2 text-xs text-white'>
+      {tokens.map((token, index) => (
+        <React.Fragment key={token}>
+          {index > 0 && <ChevronRight className='w-3 h-3 text-gray-400' />}
+          <span>{token}</span>
+        </React.Fragment>
+      ))}
```
grod220, line 27: issue:

1. It should display on hover. At the moment it is on-click.
2. When it is displaying, it seems to increase in height ever so slightly.

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+import { getSymbolFromValueView } from '@penumbra-zone/getters/value-view';
+import { usePathSymbols } from '@/pages/trade/model/use-path.ts';
 
-const RouteBookLoadingState = () => {
+import { Tabs } from '@penumbra-zone/ui/Tabs';
+
+const HopCount = ({ count }: { count: number }) => {
+  return (
+    <span className={count === 0 ? 'text-white' : 'text-[#F49C43]'}>
+      {count === 0 ? 'Direct' : `${count} ${count === 1 ? 'Hop' : 'Hops'}`}
```
grod220, line 14: issue: For direct, it should have two hops right? The first is the starting token and the second is the ending token. In this case, I think it likely should be something like:
1. Direct, if hop === 2
2. Else, hop = traces.hops.length - 1

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
-              <td className='text-right tabular-nums'>{trace.amount}</td>
-              <td className='text-right tabular-nums'>{trace.total}</td>
-              <td className='text-right tabular-nums'>{trace.hops.length}</td>
+    <div className='flex flex-col max-w-full border-y border-[#262626]'>
+      <div className='flex items-center gap-2 px-4 h-11 border-b border-[#262626]'>
+        <Tabs
+          value={'routes'}
+          onChange={() => {
+            undefined;
+          }}
```
grod220, line 92: request: can you add a second tab for `Route Depth` with text that says "Coming soon..."?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
-          {multiHops.buy.map(trace => (
-            <tr key={trace.price + trace.total} style={{ color: 'green' }}>
-              <td className='text-left tabular-nums'>{trace.price}</td>
-              <td className='text-right tabular-nums'>{trace.amount}</td>
-              <td className='text-right tabular-nums'>{trace.total}</td>
-              <td className='text-right tabular-nums'>{trace.hops.length}</td>
+    <div className='flex flex-col max-w-full border-y border-[#262626]'>
+      <div className='flex items-center gap-2 px-4 h-11 border-b border-[#262626]'>
+        <Tabs
+          value={'routes'}
```
grod220, line 89: suggestion: value should probably be called "route book" too. Wonder if we can make this type safe. Maybe an enum that is also in `ROUTEBOOK_TABS`?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
-          ))}
-          {multiHops.buy.map(trace => (
-            <tr key={trace.price + trace.total} style={{ color: 'green' }}>
-              <td className='text-left tabular-nums'>{trace.price}</td>
-              <td className='text-right tabular-nums'>{trace.amount}</td>
-              <td className='text-right tabular-nums'>{trace.total}</td>
-              <td className='text-right tabular-nums'>{trace.hops.length}</td>
+    <div className='flex flex-col max-w-full border-y border-[#262626]'>
+      <div className='flex items-center gap-2 px-4 h-11 border-b border-[#262626]'>
+        <Tabs
```
grod220, line 88: question: I don't suppose there is a prop or something for condensed is there? It feels a bit too big right?

package.json
```diff
@@ -76,6 +76,7 @@
     "react-dom": "^18.3.1",
     "react-outside-click-handler": "^1.3.0",
     "sharp": "^0.33.5",
+    "usehooks-ts": "^3.1.0",
```
grod220, line 79: question: Used anywhere?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+  trace,
+  isSell,
+  relativeSize,
+}: {
+  trace: Trace;
+  isSell: boolean;
+  relativeSize: number;
+}) => {
+  const [showRoute, setShowRoute] = useState(false);
+  const bgColor = isSell ? 'rgba(175, 38, 38, 0.24)' : 'rgba(28, 121, 63, 0.24)';
```
grod220, line 42: suggestion: maybe these strings should be constants at the top

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+          actionType='accent'
+        />
+      </div>
+
+      <div className='flex-1'>
+        <table className='w-full'>
+          <thead>
+            <tr className='text-xs text-gray-400'>
+              <th className='py-[8px] font-normal text-left'>Price({pair.quoteSymbol})</th>
+              <th className='py-[8px] font-normal text-right'>Amount({pair.baseSymbol})</th>
```
grod220, line 103: suggestion: Can we convert the px amount to a default tailwind choice if possible? I believe py-[8px] is the same as py-2 right?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+    return <div className='text-red-500'>Error loading route book: {String(bookErr)}</div>;
   }
 
-  return <RouteBookData bookData={data} />;
+  return <RouteBookData bookData={bookData} />;
 });
+
+const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
+  if (!sellOrders.length || !buyOrders.length) {
+    return null;
```
grod220, line 152: suggestion: This may be a specific Gabe-issue, but I kinda strongly feel like `null` was a mistake in the javascript language. Sometimes we are required to use it, but whenever possible think we should be using `undefined` instead.

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
-    return <RouteBookLoadingState />;
+  if (bookErr) {
+    return <div className='text-red-500'>Error loading route book: {String(bookErr)}</div>;
   }
 
-  return <RouteBookData bookData={data} />;
+  return <RouteBookData bookData={bookData} />;
 });
+
+const calculateSpread = (sellOrders: Trace[], buyOrders: Trace[]) => {
```
grod220, line 150: suggestion: This it would be helpful to pull this out into maybe the `models` folder and write a few tests for this.

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+    return null;
+  }
+
+  return (
+    <tr>
+      <td colSpan={4} className='border-y border-[#262626]'>
+        <div className='flex items-center justify-center gap-2 px-3 py-3 text-xs'>
+          <span className='text-[#55D383]'>{spreadInfo.midPrice}</span>
+          <span className='text-gray-400'>Spread:</span>
+          <span className='text-white'>{spreadInfo.amount} USDC</span>
```
grod220, line 189: issue: `USDC` should not be hardcoded right? This should probably be the quote asset symbol.

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+const SpreadRow = ({ sellOrders, buyOrders }: { sellOrders: Trace[]; buyOrders: Trace[] }) => {
+  const spreadInfo = calculateSpread(sellOrders, buyOrders);
+
+  if (!spreadInfo) {
+    return null;
+  }
+
+  return (
+    <tr>
+      <td colSpan={4} className='border-y border-[#262626]'>
```
grod220, line 185: question: `#262626` (and all the other color strings) should probably be a constant too right? It's likely we should have these constants in some global place somewhere. Maybe in the tailwind config (?) until we have the UI library updated I suppose.

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
+
+const calculateRelativeSizes = (orders: Trace[]): Map<string, number> => {
+  if (!orders.length) {
+    return new Map();
+  }
+
+  const totals = orders.map(order => parseFloat(order.total));
+  const maxTotal = Math.max(...totals);
+
+  return orders.reduce((map, order) => {
```
grod220, line 205: question: can't this be` totals.reduce()`?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
 
-  if (isLoading || !data) {
-    return <RouteBookLoadingState />;
+  if (bookErr) {
+    return <div className='text-red-500'>Error loading route book: {String(bookErr)}</div>;
   }
 
-  return <RouteBookData bookData={data} />;
+  return <RouteBookData bookData={bookData} />;
 });
```
grod220, line 148: nit: probably not necessary to rename maybe?

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,212 @@
...
 
 export const RouteBook = observer(() => {
-  const { data, isLoading, error } = useBook();
+  const { data: bookData, isLoading: bookIsLoading, error: bookErr } = useBook();
 
-  if (error) {
-    return <div>Error loading route book: ${String(error)}</div>;
+  if (bookIsLoading || !bookData) {
+    return <div className='text-gray-400'>Loading...</div>;
```
grod220, line 140: issue: if there is an error, the `!bookData` will guarantee it will be stuck in this loading state. Think the previous ordering prevented that possibility.

grod220: question: Perhaps the loading skeleton can be included too? Or if not in this PR, perhaps a subsequent PR?

package.json
vacekj: Tried using it for the on-hover routes, but it was flaky even with this hook. Will remove it if I make it work on-hover in some way. Thinking of using the ::after's 'content' rn.

src/pages/trade/ui/route-book.tsx
vacekj: Good idea, done!

src/pages/trade/ui/route-book.tsx
vacekj: Good catch - tailwind config sounds good to me for now, then directly in the UI pkg later.

src/pages/trade/ui/route-book.tsx
vacekj: indeed, fixed and optimized this.

src/pages/trade/ui/route-book.tsx
vacekj: Done and done 👍

src/pages/trade/ui/route-book.tsx
```diff
@@ -1,60 +1,166 @@
...
+                      isSell={false}
+                      relativeSize={buyRelativeSizes.get(trace.total) ?? 0}
+                    />
+                  ))}
+                </>
+              ) : (
+                Array(17)
+                  .fill(1)
+                  .map((_, i) => <SkeletonRow isSpread={i === 8} key={i} />)
+              )}
```
grod220, line 104: question: we aren't using the isLoading var anymore, are we covering all user paths? 🤔 or maybe we can rely on an `undefined` for data?

grod220 (approved):

src/pages/trade/ui/trade-row.tsx
```diff
@@ -0,0 +1,86 @@
...
+          height: 33px;
+          background: rgba(250, 250, 250, 0.05);
+        }
+      `}</style>
+
+      {/* Route display that shows on hover */}
+      <td
+        className='hidden group-hover:flex justify-center absolute left-0 right-0 px-4'
+        colSpan={4}
+        style={{ visibility: 'visible', zIndex: 30 }}
```
grod220, line 59: nit: can you make this non-selectable? Is it `cursor: none` or something?

src/pages/trade/ui/trade-row.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+        tr:hover::after {
+          content: '';
+          position: absolute;
+          left: 0;
+          right: 0;
+          height: 33px;
+          background: rgba(250, 250, 250, 0.05);
+        }
+      `}</style>
```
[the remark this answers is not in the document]
vacekj: Not something that tailwind was designer for, so can't be done without going to unergonomic lenghts I'm afraid.

src/pages/trade/ui/route-book.tsx
vacekj: undefined for data is fine, since we are using it in place of isLoading. for catching errors, we have line 120.

src/pages/trade/ui/trade-row.tsx
vacekj: Sure thing.

src/pages/trade/model/trace.ts
```diff
@@ -0,0 +1,27 @@
...
+  const buyPrice = parseFloat(highestBuy.price);
+
+  const spread = sellPrice - buyPrice;
+  const midPrice = (sellPrice + buyPrice) / 2;
+  const spreadPercentage = (spread / midPrice) * 100;
+
+  return {
+    amount: spread.toFixed(8),
+    percentage: spreadPercentage.toFixed(2),
+    midPrice: midPrice.toFixed(8),
```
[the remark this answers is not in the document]
vacekj: Good catch - fixed.

src/pages/trade/ui/trade-row.tsx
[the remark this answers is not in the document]
VanishMax: sure can:

```tsx
<tr className='h-[33px] border-b border-border-faded group [&:hover>td]:invisible hover:after:absolute hover:after:left-0 hover:after:right-0 hover:after:h-[33px] hover:after:bg-other-tonalFill5'>
```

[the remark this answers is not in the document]
VanishMax: notice the `[&:hover>td]:invisible` class

src/pages/trade/ui/trade-row.tsx
VanishMax, line 59: question: probably can be done with classes?

```suggestion
        style={{ visibility: 'visible', zIndex: 30 }}
        className="visible z-30"
```

tailwind.config.ts
```diff
@@ -1,9 +1,27 @@
...
+        sell: {
+          bg: 'rgba(175, 38, 38, 0.24)',
+          text: '#F17878',
+        },
+        buy: {
+          bg: 'rgba(28, 121, 63, 0.24)',
+          text: '#55D383',
+        },
+        route: {
+          text: '#F49C43',
```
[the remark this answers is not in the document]
vacekj: Yes - corrected that.

src/pages/trade/ui/trade-row.tsx
vacekj: Fixed.

src/pages/trade/ui/trade-row.tsx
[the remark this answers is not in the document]
VanishMax: suggestion: about the unergonomic lengths – you can wrap the className into the `cn` function and make it multiline with logical class separation

package.json
```diff
@@ -77,6 +77,7 @@
     "react": "^18.3.1",
     "react-chartjs-2": "^5.2.0",
     "react-dom": "^18.3.1",
+    "react-loading-skeleton": "^3.5.0",
```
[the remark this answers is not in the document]
VanishMax: Maybe it would be faster to do it here, without extra issues?

Issue #121, opened 2024-10-31
bug: registry assets fail to load in prod build


conorsch:
Spent some time debugging a disparity between the deployed testnet-facing instance (https://dex-explorer.testnet.plinfra.net) and the local dev env. That investigation resulted in https://github.com/penumbra-zone/dex-explorer/pull/120, which is good progress, but we're still not quite there: the v2 version now deployed doesn't load asset info correctly.

When running the prod build, I see in the browser network console a 404 when accessing this URL: `https://raw.githubusercontent.com/prax-wallet/registry/main/registry/chains/.json` That's obviously a concatenation error where the env var `PENUMBRA_CHAIN_ID=penumbra-testnet-phobos-2` did not get interpolated into the string. It's definitely set, though, and the `pnpm dev` environment loads it just fine. Why isn't the chain id loading in the prod build?

Previously we were using the bundled version of the assets (https://github.com/penumbra-zone/dex-explorer/issues/57#issuecomment-2363412022), but I believe recent work on the v2 interface moved that to remote fetching.

grod220:
Some ideas, but happy to pair on this though!

Wonder if we need to read from the build args and insert as env vars in the [containerfile](https://github.com/penumbra-zone/dex-explorer/blob/main/Containerfile):

```
ARG PENUMBRA_GRPC_ENDPOINT
ENV PENUMBRA_GRPC_ENDPOINT=${PENUMBRA_GRPC_ENDPOINT}
```

Do we need to pass in secrets within the [deploy action](https://github.com/penumbra-zone/dex-explorer/blob/main/.github/workflows/deploy.yml)?

Another idea entirely is to write a `env.production` file with all the secrets ([docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables)). Frontenders are using `env.development` locally and nextjs automatically loads the variables.

conorsch:
Ah, are you suggesting that the the code has regressed and we lost the ability to configure the asset registry info at runtime? In #73 we discussed the requirement to support multiple chains via our published image:

> We currently build a container image, ghcr.io/penumbra-zone/dex-explorer, to make it easy for anyone to run a dex-explorer. There are currently multiple deployments, tracking different chains, based on this image. We must update the application config such that the same built artifact, i.e. the container, can be used to run against multiple chains.

For example, RC runs an instance and they need to be able to configure the correct chain info. 

[...]

PR #125, opened 2024-11-04
Make env route dynamic

grod220:
Possibly closes https://github.com/penumbra-zone/dex-explorer/issues/121

By default, the `api/env` route was being built statically at build time. It looks like Nextjs recognized that it depended on static variables (env vars) and thought it a good idea to pre-build the route's response #facepalm. 

Before:
Screenshot 2024-11-04 at 10 43 03 AM

After:

Screenshot 2024-11-04 at 10 42 08 AM

### DIFF
--- a/app/api/env/route.ts
+++ b/app/api/env/route.ts
@@ -1,1 +1,4 @@
 export { GET } from '@/shared/api/server/env';
+
+// Prevents static generation of API route result at build time
+export const dynamic = 'force-dynamic';


conorsch:
Nice! This change resolved the specific problem reported in #121, where testnet asset lookup was failing. Running this new code, I see in the browser console that the correct lookup is performed via GET:

dex-successful-asset-lookup

which confirms that the server-side env var is now properly configuring the app behavior, as we wanted. Problems still remain in terms of overall app stability, but will file follow-up issues for those so they can be addressed appropriately.

conorsch (approved):

Issue #92, opened 2024-10-09
Make initial route redirect configurable


conorsch:
Filing as follow-up to #80. 

> It's a nice touch to redirect the trading view to `penumbra:gm` on the testnet; however, that route should probably be configurable, too, maybe with the previous setting of `penumbra:usdc` retained as default behavior. Otherwise, deployments against the `penumbra-1` chain will show nonsense data, because there's no `gm` asset. 

_Originally posted by @conorsch in https://github.com/penumbra-zone/dex-explorer/pull/80#pullrequestreview-2354937881_

Issue #107, opened 2024-10-24
Add design to 24hr stats


grod220:
<img width="375" alt="Screenshot 2024-10-24 at 4 34 59 PM" src="https://github.com/user-attachments/assets/0b3b852b-852d-4d14-894c-61b9e3119345">

Reference: https://www.figma.com/design/xWsVprkEgEihweVN9isrGQ/DEX-(Incl.-UI-Lib)?node-id=3658-145100&node-type=canvas&t=NYogohDveZXB9V4J-0

cronokirby:
https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pindexer/src/dex_ex/schema.sql#L50

I think the dex explorer schema in pindexer should already be able to provide what's needed here.

PR #128, opened 2024-11-05
Candles v2 (time windows)

grod220:
Closes https://github.com/prax-wallet/prax/issues/229
Closes https://github.com/penumbra-zone/dex-explorer/issues/92

- Adds new db query to get window-based candles from new db
- Removes old candle route + dependency on old events-based db
- Assortment of refactoring
- Note the design is not quite like the figma. Follow up UI work will be needed.

https://github.com/user-attachments/assets/6398666e-72d7-4e2a-8622-2680faef5aef


src/pages/trade/api/book.ts
```diff
@@ -1,27 +1,28 @@
...
 import { useRefetchOnNewBlock } from '@/shared/api/compact-block.ts';
-import { RouteBookResponse, RouteBookResponseJson } from '@/shared/api/server/book/types';
+import { RouteBookResponse } from '@/shared/api/server/book/types';
 import { deserializeRouteBookResponseJson } from '@/shared/api/server/book/serialization.ts';
+import { RouteBookApiResponse } from '@/shared/api/server/book';
+import { usePathSymbols } from '@/pages/trade/model/use-path.ts';
 
-export const useBook = (symbol1: string | undefined, symbol2: string | undefined) => {
+export const useBook = () => {
+  const { baseSymbol, quoteSymbol } = usePathSymbols();
```
grod220, line 9: Realizing a few hooks are relying upon async data fetching for the path symbols. This is not necessary because they are synchronously available. So created a new `usePathSymbols();` for that.

src/pages/trade/api/book.ts
```diff
@@ -1,27 +1,28 @@
...
       };
       const baseUrl = '/api/book';
       const urlParams = new URLSearchParams(paramsObj).toString();
       const res = await fetch(`${baseUrl}?${urlParams}`);
-      const data = (await res.json()) as RouteBookResponseJson;
-      return deserializeRouteBookResponseJson(data);
+      const jsonRes = (await res.json()) as RouteBookApiResponse;
+      if ('error' in jsonRes) {
+        throw new Error(jsonRes.error);
+      }
```
grod220, line 23: Because nextjs can return an error field, we need to parse it and re-throw if we see it

src/pages/trade/redirect.tsx
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  if (error) {
+    return <div className='text-red-600'>{String(error)}</div>;
+  } else if (isLoading || !data) {
+    return <div className='text-white'>Loading...</div>;
+  } else {
+    redirect(`/trade/${data.baseAsset}/${data.quoteAsset}`);
+  }
+};
```
grod220, line 38: Was finding it a bit annoying that we had a hard-coded pair here. This causes issues when going back and forth between testnet/mainnet. When ahead and created a hook to pull the highest priority pair and show that as the default.

src/shared/api/indexer/connector.tsx
```diff
@@ -1,116 +0,0 @@
-import { Pool, QueryConfigValues, QueryResultRow } from 'pg';
-import fs from 'fs';
-import { BlockInfo } from './lps';
-import { JsonValue } from '@bufbuild/protobuf';
-
-const INDEXER_CERT = process.env['PENUMBRA_INDEXER_CA_CERT'];
-
-// TODO: Delete when possible, this is deprecated
-export class IndexerQuerier {
-  private pool: Pool;
```
grod220, line 10: Bye bye old db client

src/shared/api/server/candles/utils.ts
grod220, line 1: Nearly this whole file was copied over from the previous implementation

src/shared/database/index.ts
```diff
@@ -41,6 +41,25 @@ class Pindexer {
...
+    return this.db
+      .selectFrom('dex_ex_price_charts')
+      .innerJoin('dex_ex_candlesticks', 'candlestick_id', 'dex_ex_candlesticks.id')
+      .select(['start_time', 'open', 'close', 'low', 'high', 'swap_volume', 'direct_volume'])
+      .where('the_window', '=', window)
+      .where('asset_start', '=', Buffer.from(baseAsset.inner))
+      .where('asset_end', '=', Buffer.from(quoteAsset.inner))
+      .where('start_time', '<', start)
+      .where('start_time', '>=', end)
+      .execute();
```
grod220, line 61: 3177e7a6d5cdaa9c35be1f0dd6a2c236
@cronokirby

grod220:
@conorsch after this is merged, the nextjs app will no longer have a dependency on the old database and only on the new one fyi.

conorsch (changes requested):
Encountered one large bug related to the initial route redirect, see screenshot shared above. Standing by to assist with backend changes once we're happy with the diff.

src/pages/trade/api/candles.tsx
```diff
@@ -1,21 +1,37 @@
 import { useQuery } from '@tanstack/react-query';
-import { CandlestickData } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
 import { useRefetchOnNewBlock } from '@/shared/api/compact-block.ts';
+import { CandleApiResponse } from '@/shared/api/server/candles/types.ts';
+import { usePathSymbols } from '@/pages/trade/model/use-path.ts';
+import { OhlcData } from 'lightweight-charts';
+import { DurationWindow } from '@/shared/database/schema.ts';
+
+const DEX_ENABLED_DATE = '2024-08-01';
```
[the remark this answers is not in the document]
grod220: Good catch. We should make this chain agnostic. Given we have such little data, it's hard for me to determine what is the upper bound for this query. Aka, should we only be pulling data from the last year? Alas, I'm going to remove the start/end date logic and just pull whatever candlestick data is in the database. Think it's worthwhile revisiting this topic in a year.

src/pages/trade/redirect.tsx
grod220: Definitely a bug in the implementation. I should have been filtering out delegation tokens, fixed! Also, added some sample priority scores: https://github.com/prax-wallet/registry/pull/109.

conorsch: Nice, confirming resolved on my end:

dex-explorer-getting-better

Looks like the styling is sub-optimal but we can handle that in time.

grod220: Indeed, but we have an issue for that: https://github.com/penumbra-zone/dex-explorer/issues/107

grod220:
(@conorsch OOO today, but gave the approval ✅ offline)

PR #133, opened 2024-11-14
feat: #110: Explore page

### LINKED ISSUE #110: Explore page header
Desktop Skeleton:

Screenshot 2024-10-24 at 4 55 30 PM

Desktop Loaded state:
Screenshot 2024-10-24 at 4 55 50 PM

Mobile skeleton:

Screenshot 2024-10-24 at 4 56 29 PM

Mobile Loaded state:

Screenshot 2024-10-24 at 4 56 45 PM

Add penumbra background image as well 🌔 

### LINKED ISSUE #111: Explore page: trading pairs
Desktop loading skeleton:

Screenshot 2024-10-24 at 4 59 40 PM

Desktop loaded state:
Screenshot 2024-10-24 at 5 00 12 PM

Mobile loading skeleton:
Screenshot 2024-10-24 at 5 00 37 PM

Mobile loaded state:
Screenshot 2024-10-24 at 5 01 25 PM

VanishMax:
Closes #110 

The header is ready, trading pair block is only html/css ready. The data is requested and rendered server-side only – all skeletons are removed from the header.

image


src/pages/explore/ui/stats.tsx
```diff
@@ -0,0 +1,81 @@
...
+      </Text>
+    );
+  }
+
+  return (
+    <div className='grid grid-cols-1 desktop:grid-cols-3 gap-2'>
+      <InfoCard title='Total Trading Volume (24h)'>
+        <Text large color='text.primary'>
+          <Text large color='text.primary'>
+            {shortify(Number(getFormattedAmtFromValueView(stats.directVolume)))}
```
VanishMax, line 25: @grod220 the volume comes a 10^21 number, and even the `ValueViewComponent` displays it as `1589T`. Something seems to be wrong in these calculations. Do you know what?

src/pages/explore/api/get-stats.ts
```diff
@@ -0,0 +1,97 @@
...
+      return { error: `No stats found` };
+    }
+
+    // TODO: Add getMetadataBySymbol() helper to registry npm package
+    const allAssets = registry.getAllAssets();
+    // TODO: what asset should be used here?
+    const usdcMetadata = allAssets.find(asset => asset.symbol.toLowerCase() === 'usdc');
+    if (!usdcMetadata) {
+      return { error: 'USDC not found in registry' };
+    }
```
VanishMax, line 46: question: the database doesn't seem to have the base asset for the global summary. How ok is it to have this check in code?

src/pages/explore/api/get-stats.ts
```diff
@@ -0,0 +1,97 @@
...
+        largestPairEnd &&
+        toValueView({
+          amount: stats.largest_dv_trading_pair_volume,
+          metadata: largestPairEnd,
+        }),
+      liquidity: toValueView({
+        amount: parseInt(`${stats.liquidity}`),
+        metadata: usdcMetadata,
+      }),
+      directVolume: toValueView({ amount: stats.direct_volume, metadata: usdcMetadata }),
```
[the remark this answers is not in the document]
VanishMax: but it doesn't send this over the wire since everything is rendered in react server components with `use server` (or without `use client`, to be precise). The data is fetched and immediately transformed to HTML. This valid HTML is then sent over the wire to the client.

next.config.mjs
```diff
@@ -41,7 +41,26 @@ const nextConfig = {
...
+          options: {
+            svgo: false,
+            svgoConfig: {
+              plugins: [
+                {
+                  name: 'preset-default',
+                  params: {
+                    overrides: {
+                      removeViewBox: false,
+                    },
```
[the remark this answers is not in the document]
VanishMax: Allows the import of `.svg` files as react components. I used it to import PenumbraWaves component.

The `removeViewBox: false` is needed to keep the dimensions of the svg and scale it without cutting it

src/pages/explore/ui/stats.tsx
VanishMax: Would also highlight that the `liquidity` field returns something like `liquidity: 118364618673.38`, which is not an integer. It requires us to `parseInt()`, which may fail with big numbers

src/pages/explore/ui/stats.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </>
+        ) : (
+          <Text large color='text.primary'>
+            -
+          </Text>
+        )}
+      </InfoCard>
+      <InfoCard title='Total Liquidity Available'>
+        <Text large color='text.primary'>
+          {shortify(Number(getFormattedAmtFromValueView(stats.liquidity)))}
```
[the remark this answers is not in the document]
VanishMax: I tried but it there is currently no way to hide the asset icon, and the font sizes differ from the designs

src/pages/explore/ui/stats.tsx
```diff
@@ -0,0 +1,81 @@
+import { round } from '@penumbra-zone/types/round';
+import { Text } from '@penumbra-zone/ui/Text';
+import { InfoCard } from './info-card';
+import { pluralizeAndShortify } from '@/shared/utils/pluralize';
+import { shortify } from '@penumbra-zone/types/shortify';
+import { getFormattedAmtFromValueView } from '@penumbra-zone/types/value-view';
+import { getStats } from '../api/get-stats';
+
+export const ExploreStats = async () => {
```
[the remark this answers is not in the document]
VanishMax: good call! Added this layout for tablet screens

src/pages/explore/api/get-stats.ts
cronokirby: Yeah we can add metadata for this

src/pages/explore/ui/stats.tsx
cronokirby: For mainnet data, I'm only seeing weird values for the month window in the aggregate summary, will investigate, but probably related to needing some more stringent filtering of assets with no real quote asset price on penumbra.

The liquidity field could be amended to not be a float, but every other value that is conceptually an Amount (like volume) is already a float, and so I think I would recommend just developing a bit of infrastructure here to deal with that. We need bespoke code here to deal with prices, which have to be a float, as well.

Issue #132, opened 2024-11-13
db: new schema for dex events, `dex_ex_candlesticks` no longer exists


conorsch:
The `pindexer` database schema no longer includes a `dex_ex_candlesticks` table, as of merge of https://github.com/penumbra-zone/penumbra/pull/4917. The dex-explorer database interface should be updated to match the latest schema. 

Also affected are queries against the `dex_ex_summary` table, which was also removed (#126).

See related discussion for motivation in https://github.com/penumbra-zone/penumbra/issues/4914

VanishMax:
@conorsch when will the new tables be applied to pindexer? For now `dex_ex_summary` still seems to exist

PR #135, opened 2024-11-18
Update to latest pindexer schema

grod220:
Closes https://github.com/penumbra-zone/dex-explorer/issues/132

Mainnet pindexer support TBD.


VanishMax (commented):
suggestion: add the generation command to `package.json`, so it can be easily repeated by any DEX developer. You would probably need to add `kysely-codegen` to dev dependencies

src/shared/database/schema.ts
```diff
@@ -20,43 +20,71 @@ export type JsonValue = JsonArray | JsonObject | JsonPrimitive;
...
-  id: Generated<number>;
-  low: number;
-  open: number;
-  swap_volume: number;
-}
-
 export const durationWindows = ['1m', '15m', '1h', '4h', '1d', '1w', '1mo'] as const;
 export type DurationWindow = (typeof durationWindows)[number];
 export const isDurationWindow = (str: string): str is DurationWindow =>
   durationWindows.includes(str as DurationWindow);
```
VanishMax, line 26: question: are these constants and functions also generated by kysely? If not, can you move them somewhere else (e.g. `src/shared/utils`)?

src/shared/database/schema.ts
```diff
@@ -7,56 +13,259 @@ export type Generated<T> =
...
+  supply_validators: SupplyValidators;
 }
+
+export type DB = Pick<
+  RawDB,
+  | 'dex_ex_aggregate_summary'
+  | 'dex_ex_pairs_block_snapshot'
+  | 'dex_ex_pairs_summary'
+  | 'dex_ex_price_charts'
+>;
```
grod220, line 271: Previously relevant: https://github.com/penumbra-zone/dex-explorer/pull/123#pullrequestreview-2410934582 cc @cronokirby

We can't both have an auto schema generator and not include some irrelevant types. Hoping with `scripts/generate-pindexer-schema.ts` and this `Pick` it is a fair compromise given developers will not be able to query outside this list.

VanishMax (approved):
praise for the script covering all cases 👍

PR #137, opened 2024-11-18
Adjust db query for summaries

grod220:
Small follow ups from: https://github.com/penumbra-zone/dex-explorer/pull/135

### DIFF
--- a/src/pages/trade/ui/summary.tsx
+++ b/src/pages/trade/ui/summary.tsx
@@ -48,16 +48,18 @@ export const Summary = () => {
   const { quoteAsset } = usePathToMetadata();
 
   const change24h = data && {
-    positive: data.price >= data.price,
-    change: round(data.price - data.price, 4),
-    percent: round(Math.abs(((data.price - data.price) / data.price) * 100), 2),
+    positive: data.price >= data.price_then,
+    change: round(data.price - data.price_then, 4),
+    percent: !data.price
+      ? '0'
+      : round(Math.abs(((data.price - data.price_then) / data.price_then) * 100), 2),
   };
 
   if (error) {
     return (
-      <SummaryCard title='Error'>
+      <SummaryCard title=''>
         <Text detail color='destructive.light'>
-          Error: {String(error)}
+          {String(error)}
         </Text>
       </SummaryCard>
     );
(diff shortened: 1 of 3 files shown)


src/pages/trade/ui/summary.tsx
```diff
@@ -48,16 +48,18 @@ export const Summary = () => {
   const { quoteAsset } = usePathToMetadata();
 
   const change24h = data && {
-    positive: data.price >= data.price,
-    change: round(data.price - data.price, 4),
-    percent: round(Math.abs(((data.price - data.price) / data.price) * 100), 2),
+    positive: data.price >= data.price_then,
+    change: round(data.price - data.price_then, 4),
```
grod220, line 52: Bug fix, was not using `price_then` fields

src/pages/trade/ui/summary.tsx
```diff
@@ -48,16 +48,18 @@ export const Summary = () => {
...
+    percent: !data.price
+      ? '0'
+      : round(Math.abs(((data.price - data.price_then) / data.price_then) * 100), 2),
```
grod220, line 55: On testnet, prices can be 0. In that case was rendering a `NaN`.

src/pages/trade/ui/summary.tsx
```diff
@@ -48,16 +48,18 @@ export const Summary = () => {
...
   };
 
   if (error) {
     return (
-      <SummaryCard title='Error'>
+      <SummaryCard title=''>
```
grod220, line 60: The Error title was a bit redundant. And we'll need to update this to the error design anyway in figma later. 

Screenshot 2024-11-18 at 7 56 04 PM

src/pages/trade/ui/summary.tsx
```diff
@@ -96,18 +98,16 @@ export const Summary = () => {
...
       <SummaryCard title='24h Low' loading={isLoading}>
-        {data?.low && (
-          <Text detail color='text.primary'>
-            {round(data.low, 4)}
-          </Text>
-        )}
+        {/*  TODO: After added to DB, show here */}
+        <Text detail color='text.primary'>
+          -
+        </Text>
```
grod220, line 110: Screenshot 2024-11-18 at 7 55 58 PM

src/shared/database/index.ts
```diff
@@ -37,14 +37,11 @@ class Pindexer {
 
   async summary(window: DurationWindow, baseAsset: AssetId, quoteAsset: AssetId) {
     return this.db
-      .selectFrom('dex_ex_pairs_summary as s')
-      .leftJoin('dex_ex_price_charts as c', join =>
-        join.onRef('c.asset_start', '=', 's.asset_start').onRef('c.asset_end', '=', 's.asset_end'),
```
grod220, line 42: The low/high values in the charts table does not map to the summary quite as expected:

> and there's a difference between the 1d candlestick, which is "what has happened from midnight UTC until now, today" vs "what was the lowest price between now and 24h ago" - @cronokirby 

Will simplify the query and update the schema when low/high is available in the summary table

PR #139, opened 2024-11-20
Summary value conversions

grod220:
The data coming out of Pindexer is always denominated in the base asset. This means we need to convert it using the display exponent to make the result legible to the user.

The PR updates packages and adds support for this new math with the `Summary` component and API routes. 

Screenshot 2024-11-20 at 5 57 49 PM


src/pages/trade/model/trace.ts
```diff
@@ -1,5 +1,5 @@
 import { Trace } from '@/shared/api/server/book/types.ts';
-import { round } from '@/shared/utils/numbers/round.ts';
+import { round } from '@penumbra-zone/types/round';
```
grod220, line 2: Moved `round` util to `@penumbra-zone/types` as it was needed in the UI component. Also augmented there with more features and tests.

src/pages/trade/ui/summary.tsx
```diff
@@ -45,15 +46,6 @@ const SummaryCard = ({
...
   const { data, isLoading, error } = useSummary('1d');
-  const { quoteAsset } = usePathToMetadata();
-
-  const change24h = data && {
-    positive: data.price >= data.price_then,
-    change: round(data.price - data.price_then, 4),
-    percent: !data.price
-      ? '0'
-      : round(Math.abs(((data.price - data.price_then) / data.price_then) * 100), 2),
-  };
```
grod220, line 56: Coming from backend now

src/pages/trade/ui/summary.tsx
```diff
@@ -68,57 +60,82 @@ export const Summary = () => {
...
+        <Text detail color='text.primary'>
+          {data && 'price' in data ? roundTrim(data.price) : '-'}
+        </Text>
+      </SummaryCard>
+      <SummaryCard title='24h Change' loading={isLoading}>
+        {data && 'noData' in data && (
           <Text detail color='text.primary'>
-            {round(data.price, 6)}
+            -
           </Text>
```
grod220, line 71: If no summary is available defaulting to dashes for the value instead of the error message

src/shared/api/server/summary.ts
```diff
@@ -56,11 +57,43 @@ export async function GET(req: NextRequest): Promise<NextResponse<SummaryRespons
 
   const summary = results[0];
   if (!summary) {
-    return NextResponse.json(
-      { error: `No summary found for ${baseAssetSymbol}/${quoteAssetSymbol}` },
-      { status: 400 },
-    );
+    return NextResponse.json({ window: durationWindow, noData: true });
```
grod220, line 60: Because it's so common to have no summary data, created a 200 success with another payload

src/shared/api/server/types.ts
```diff
@@ -0,0 +1,79 @@
...
+      high: this.high,
+      low: this.low,
+      change: this.change,
+    };
+  }
+
+  static fromJson(json: SummaryDataResponseJson): SummaryDataResponse {
+    return new SummaryDataResponse({
+      window: json.window,
+      directVolume: ValueView.fromJson(json.directVolume),
```
grod220, line 65: Because we are passing a `ValueView` over the wire, we need to add serializaiton/deserialization logic to the return type

src/shared/api/server/summary.ts
```diff
@@ -56,11 +57,43 @@ export async function GET(req: NextRequest): Promise<NextResponse<SummaryRespons
...
+            decimals: 2,
+          }),
+  };
+
+  const dataResponse = new SummaryDataResponse({
+    window: durationWindow,
+    directVolume: toValueView({
+      amount: summary.direct_volume_over_window,
+      metadata: quoteAssetMetadata,
+    }),
```
grod220, line 81: Passing volume as a `ValueView` allows us to use the normal `ValueViewComponent` that handles the display exponents correctly automatically.

src/pages/trade/ui/summary.tsx
```diff
@@ -68,57 +60,82 @@ export const Summary = () => {
...
+
+const getColor = (res: SummaryDataResponse, version: 'text' | 'bg') => {
+  if (res.change.sign === 'positive') {
+    return `${version}-success-light`;
+  }
+  if (res.change.sign === 'negative') {
+    return `${version}-destructive-light`;
+  }
+  return `${version}-neutral-light`;
+};
```
VanishMax, line 141: issue(not-so-blocking): this can only work if there's another component on the page with the same class defined statically. Would be better to extend the function to provide static classes to avoid the bug.

```suggestion
const getColor = (res: SummaryDataResponse, isBg = false): string => {
  if (res.change.sign === 'positive') {
    return isBg ? 'bg-success-light' : 'text-success-light';
  }
  if (res.change.sign === 'negative') {
    return isBg ? 'bg-destructive-light' : 'text-destructive-light';
  }
  return isBg ? 'bg-neutral-light' : 'text-neutral-light';
};
```

VanishMax (approved):

src/pages/trade/ui/summary.tsx
grod220: Ah, TIL more about the tailwind compiler. Will update 👍

PR #142, opened 2024-11-20
feat: mobile view of portfolio page

### LINKED ISSUE #109: Detect mobile on portfolio page
If a user is trying to visit the portfolio page, we should have a hook that detects mobile and renders the following view if detected. See [getCompatibility()](https://github.com/penumbra-zone/web/blob/074584500b1dd143c927cadb8ceb85d0b72e5204/packages/ui/components/ui/incompatible-browser-banner/get-compatibility.ts#L16) in minifront as reference for parsing user-agent.

Screenshot 2024-10-24 at 4 41 00 PM

vacekj:
fixes #109

### DIFF
--- a/app/portfolio/page.tsx
+++ b/app/portfolio/page.tsx
@@ -1,0 +1,9 @@
+import { PortfolioPage } from '@/pages/portfolio';
+import { headers } from 'next/headers';
+import { userAgent } from 'next/server';
+
+export default function Portfolio() {
+  const headersList = headers();
+  const { device } = userAgent({ headers: headersList });
+  return <PortfolioPage isMobile={device.type === 'mobile'} />;
+}
--- a/src/pages/portfolio/index.tsx
+++ b/src/pages/portfolio/index.tsx
@@ -1,11 +1,49 @@
 'use client';
 
+import React from 'react';
+import { XCircle } from 'lucide-react';
+import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
+import { Density } from '@penumbra-zone/ui/Density';
+import mobile from 'is-mobile';
 
 export const PortfolioPage = () => {
-  return (
-    <section>
-      <Text h2>Hi!</Text>
+  const isOnMobile = mobile();
+
+  return isOnMobile ? (
+    <section className='absolute inset-0 flex flex-col items-center justify-between p-4 gap-3 border-t  border-neutral-800'>
+      <div className='flex flex-col justify-center items-center p-0 gap-4 w-full flex-grow'>
+        <div className='relative'>
+          <XCircle className='text-neutral-400  w-8 h-8' />
+        </div>
+
+        <Text color={'text.secondary'} align={'center'} small={true}>
+          This page requires a connection to your wallet, please switch to a desktop device.
+        </Text>
+
+        <Density compact={true}>
+          {/* Copy Link Button */}
+          <Button
+            onClick={() => {
+              // We discard the promise using void,
+              // because Button only expects void-returning functions.
+              void (async () => {
+                const currentUrl = window.location.href;
+                await navigator.clipboard.writeText(currentUrl);
+              })();
+            }}
+          >
+            Copy Link
+          </Button>
+        </Density>
+      </div>
+
+      {/* Go Back Button */}
+      <Button>
+        <span className="text-sm font-medium text-white font-['Poppins']">Go Back</span>
+      </Button>
     </section>
+  ) : (
+    'Hi from desktop!'
   );
 };
--- a/src/pages/portfolio/index.tsx
+++ b/src/pages/portfolio/index.tsx
@@ -6,8 +6,16 @@ import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
 import { Density } from '@penumbra-zone/ui/Density';
 
-export const PortfolioPage = (props: { isMobile: boolean }) => {
-  return props.isMobile ? (
+interface PortfolioPageProps {
+  isMobile: boolean;
+}
+
+export function PortfolioPage({ isMobile }: PortfolioPageProps): React.ReactNode {
+  return isMobile ? <MobilePortfolioPage /> : <DesktopPortfolioPage />;
+}
+
+function MobilePortfolioPage() {
+  return (
     <section className='absolute inset-0 h-screen flex flex-col items-center justify-between p-4 gap-3 border-t border-neutral-800'>
       <div className='flex flex-col justify-center items-center p-0 gap-4 w-full flex-grow'>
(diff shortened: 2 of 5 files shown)


src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,49 @@
...
+import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
+import { Density } from '@penumbra-zone/ui/Density';
+import mobile from 'is-mobile';
 
 export const PortfolioPage = () => {
-  return (
-    <section>
-      <Text h2>Hi!</Text>
+  const isOnMobile = mobile();
```
VanishMax, line 11: issue (blocking): this `mobile()` call results in the hydration failure because this code doesn't run on the server. To render it correctly on the server, the library proposes passing the nextjs' `req` object as the function argument.

image

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,49 @@
...
+
+  return isOnMobile ? (
+    <section className='absolute inset-0 flex flex-col items-center justify-between p-4 gap-3 border-t  border-neutral-800'>
```
VanishMax, line 14: issue (ui): the section renders as follows on my macbook, not matching the designs

image

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,49 @@
...
+      <div className='flex flex-col justify-center items-center p-0 gap-4 w-full flex-grow'>
```
VanishMax, line 15: suggestion: can you please reuse the same block and apply it to the mobile sidebar (opened from header)? https://www.figma.com/design/xWsVprkEgEihweVN9isrGQ/DEX-(Incl.-UI-Lib)?node-id=3663-143558&node-type=frame&m=dev

this is not blocking but might be easier to do it right here than to open one more issue/pr

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,49 @@
...
+        <div className='relative'>
+          <XCircle className='text-neutral-400  w-8 h-8' />
```
VanishMax, line 17: suggestion (namings): use here and everywhere else the named color classes, e.g. `text-neutral-light` instead of `text-neutral-400`. This might actually be easier knowing that Figma shows the named colors next to its usage:

Screenshot 2024-11-21 at 10 33 33

VanishMax: nitpick: those double spaces in classNames make my eyes twitch 😂

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,49 @@
...
+            }}
+          >
+            Copy Link
+          </Button>
+        </Density>
+      </div>
+
+      {/* Go Back Button */}
+      <Button>
+        <span className="text-sm font-medium text-white font-['Poppins']">Go Back</span>
```
VanishMax, line 43: suggestion:

```suggestion
        <Text body>Go Back</span>
```

VanishMax (changes requested):

src/pages/portfolio/index.tsx
vacekj: huh, weird. will fix.

src/pages/portfolio/index.tsx
vacekj: sorry, will fix :D

src/pages/portfolio/index.tsx
vacekj: We can't get the req object easily because we are using /pages, not /app. Using useEffect to get around this.

src/pages/portfolio/index.tsx
vacekj: I'm not sure what you mean here?

src/pages/portfolio/index.tsx
VanishMax: Maybe using the Nextjs' `userAgent` function in the `middleware` might be a solution? https://nextjs.org/docs/app/api-reference/functions/userAgent

vacekj: I don't believe this would help us - this augments the request, but we still can't get the req object, without modifying the structure of the app. Not worth it imo, useEffect is just fine.

vacekj: Fixed this, really happy with the result now.

vacekj: LMK if it works for you.

app/portfolio/page.tsx
```diff
@@ -0,0 +1,9 @@
+import { PortfolioPage } from '@/pages/portfolio';
+import { headers } from 'next/headers';
+import { userAgent } from 'next/server';
+
+export default function Portfolio() {
+  const headersList = headers();
+  const { device } = userAgent({ headers: headersList });
+  return <PortfolioPage isMobile={device.type === 'mobile'} />;
+}
```
grod220, line 9: This is a good approach because we get SSR rendering performance benefits

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,46 @@
...
+import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
+import { Density } from '@penumbra-zone/ui/Density';
 
-export const PortfolioPage = () => {
-  return (
-    <section>
-      <Text h2>Hi!</Text>
+export const PortfolioPage = (props: { isMobile: boolean }) => {
+  return props.isMobile ? (
```
grod220, line 10: issue: instead of a large ternary can you make two components? Maybe `<MobilePortfolioPage />` & `<DesktopPortfolioPage />`?

grod220 (approved):

src/pages/portfolio/index.tsx
```diff
@@ -1,11 +1,46 @@
...
+        <Density compact={true}>
+          {/* Copy Link Button */}
+          <Button
+            onClick={() => {
+              // We discard the promise using void,
+              // because Button only expects void-returning functions.
+              void (async () => {
+                const currentUrl = window.location.href;
+                await navigator.clipboard.writeText(currentUrl);
+              })();
```
grod220, line 30: issue: think this not the way nextjs does navigation: https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating

vacekj: This just writes the url to clipboard, the `navigator` here is a red herring 😃

vacekj: added a comment so that it's clear.

src/pages/portfolio/index.tsx
vacekj: done

VanishMax (approved):
Wow, that's extremely simple! 🔥

PR #145, opened 2024-11-21
feat: #143: experimentally request summary data on server

### LINKED ISSUE #143: Fetch and render critical data on the server
Right now, `useQuery` always runs on the client and `isLoading` on server. Even though we have beautiful Skeleton components, it is not a desirable behavior. 

Ultimately we want to render the data on the server. For example, the summary row with price, volume and other important data is considered critical since it might be used not only by users but also by search bots. They rely on server returning a html string with all data in there. 

Speaking of performance, rendering data on the server might slow down the First Contentful Paint (FCP) but improve all other metrics, overall improving the performance score.

References to React Query SSR implementation: 
- https://tanstack.com/query/v5/docs/framework/react/guides/ssr
- https://tanstack.com/query/v5/docs/framework/react/guides/advanced-ssr
- https://tanstack.com/query/v5/docs/framework/react/examples/nextjs-app-prefetching

That will not be easy. We would have to use the code written for `app/api` directory inline in the server components, possibly duplicating react-query code; to deal with the serialization issues; to just learning how it works in-depth and transferring the knowledge to the team.

VanishMax:
Fixes #143 

⚠️ EXPERIMENTAL

This is a draft of moving data request on server. So far it looks like too much work. All of the changed code only moves the `summary` request to server components. This is how it works:
1. When a user types the URL, the browser then sends the URL to our server
2. The server tries to pre-render the HTML string of the server components
3. It makes the request to the `summary` 
4. The result of this request is used to render the html without `isLoading` in react query. If user has JavaScript disabled in their browser, it would render the html of the summary block normally, without any skeleton loaders
5. The serialized data of the request is also sent to the browser via react-query's `HydrationBoundary` component. See the screenshot below – it shows how react-query populates the html with the server-side data that is then taken by the client
6. When the client component `Summary` tries to `useQuery`, it takes the data from the server instead of making the new request

Now, what do you think of this approach? Should we continue investigating the best pattern of server-side request?

Screenshot 2024-11-21 at 16 26 13

### DIFF
--- a/app/app.tsx
+++ b/app/app.tsx
@@ -6,7 +6,7 @@ import { QueryClientProvider } from '@tanstack/react-query';
 import { ToastProvider } from '@penumbra-zone/ui/Toast';
 import { TooltipProvider } from '@penumbra-zone/ui/Tooltip';
 import { Header, SyncBar } from '@/widgets/header';
-import { queryClient } from '@/shared/const/queryClient';
+import { getQueryClient } from '@/shared/const/queryClient';
 
 // Used so that observer() won't subscribe to any observables used in an SSR environment
 // and no garbage collection problems are introduced.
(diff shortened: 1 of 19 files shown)


TalDerei:
@VanishMax should we close this as outdated?

Issue #151, opened 2024-11-22
Merge candle logic update


grod220:
@cronokirby to update pindexer to handle candlestick merging. The API route can remove all merge logic and rely on the single-direction query from the database.

Secondly, the backend should pass base-denominated values back to the frontend who is responsible for doing the base->display conversions.

cronokirby:
I think on the dex explorer side the change should just be to only make one database query, and pindexer itself should worry about making the chart have both pieces of information. On the dex ex side we should just need the frontend to fill in empty candles (on the frontend, to not transfer that data)

grod220:
Ah, in that case, will revise the description and please ping this issue when the merging can be removed from the API please 🙏

Issue #148, opened 2024-11-21
Populate empty candles


grod220:
From Pindexer, a candlestick event will not be emitted when there's no activity. The nextjs candlestick api route should handle filling in these gaps correctly.

With the bounds from oldest to most recent, we should look at the timestamps for each candle and create empty candlesticks to fill it in the missing events in the response object.

PR #159, opened 2024-11-22
candles: rely on database for symmetrization, fill empty candles

cronokirby:
Closes https://github.com/penumbra-zone/dex-explorer/issues/148
Closes https://github.com/penumbra-zone/dex-explorer/issues/151

This changes the logic to rely on the database to integrate price information for both directions, which simplifies the logic on our end, and makes us correctly integrate tweaks to this logic without needing changes.

This also fills empty candles in between the start and end dates of the candles provided by the database.


grod220 (approved):
Very nice!

Issue #152, opened 2024-11-22
Remove absolute amount from summary


grod220:
Percent likely does not need to be in a pill any longer

<img width="136" alt="Screenshot 2024-11-22 at 1 01 31 PM" src="https://github.com/user-attachments/assets/93e3cc27-4cdf-4b41-85f4-5241334ec9ad">

PR #164, opened 2024-11-25
Remove change value

grod220:
Closes https://github.com/penumbra-zone/dex-explorer/issues/152

Before:
Screenshot 2024-11-25 at 1 08 11 PM

After:
Screenshot 2024-11-25 at 1 08 05 PM

### DIFF
--- a/src/pages/trade/ui/summary.tsx
+++ b/src/pages/trade/ui/summary.tsx
@@ -1,5 +1,4 @@
 import { ReactNode } from 'react';
-import cn from 'clsx';
 import { useAutoAnimate } from '@formkit/auto-animate/react';
 import { Text } from '@penumbra-zone/ui/Text';
 import { Skeleton } from '@/shared/ui/skeleton';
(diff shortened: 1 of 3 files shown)


VanishMax (commented):
question: what is the reason of removing the change value? It was really useful for me personally, it is almost the first thing I look for in similar websites

grod220:
Amongst the feedback in the review from @hdevalence. I'm not sure a base change number denominated in the asset denom is that understandable by users.

hdevalence:
Yeah, open to discussion on it but my opinion is/was that reporting an absolute number for the change is less informative than a percent and that it's visually cleaner to just show the relative number. But if others feel that it's useful we can leave it in

hdevalence:
I have weak conviction on this point

TalDerei:
I personally find it very useful to keep, as it's the first thing I look for as well. Dolemite only shows the percentage change, but I found myself spending more time trying to interpret the price information.

Screenshot 2024-11-25 at 6 58 19 AM

PR #168, opened 2024-11-27
Display positions [part 1]

grod220:
First part of https://github.com/prax-wallet/prax/issues/237

Fetch + rendering logic for liquidity positions. Follow up needed to represent other fields and finish css. 

Screenshot 2024-11-27 at 1 39 33 PM

### DIFF
--- a/src/pages/trade/ui/positions.tsx
+++ b/src/pages/trade/ui/positions.tsx
@@ -1,41 +1,103 @@
-import { HeaderCell, LoadingCell } from './market-trades';
+'use client';
+
+import { Cell, HeaderCell, LoadingCell } from './market-trades';
+import { connectionStore } from '@/shared/model/connection';
+import { observer } from 'mobx-react-lite';
+import { Text } from '@penumbra-zone/ui/Text';
+import { usePositions } from '@/shared/api/positions.ts';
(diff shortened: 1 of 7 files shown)


src/pages/trade/ui/positions.tsx
```diff
@@ -1,41 +1,121 @@
...
+                <Text detail color='text.secondary'>
+                  -
+                </Text>
+              </Cell>
+              <Cell>
+                <ValueViewComponent valueView={p.asset1} context='table' />
+              </Cell>
+              <Cell>
+                <ValueViewComponent valueView={p.asset2} context='table' />
+              </Cell>
```
grod220, line 89: This logic needs to be amended in the next PR

src/shared/api/positions.ts
VanishMax, line 1: suggestion(non-blocking): move this file to `src/pages/trade/api`, if it isn't used anywhere else for not and if it isn't expected to be used somewhere else but trade page

VanishMax (approved):

hdevalence:
Looks like a good start. I think it would be good to prioritize closing the loop on the LP actions (the close+withdraw flow) before figuring out the rest of the rendering. It's easier to tweak the rendering later but that unlocks actually creating and modifying positions.

PR #170, opened 2024-11-29
feat: #111: trading pairs part 1

VanishMax:
References #111 

Part 1: requesting and displaying table data for pairs in the explore page

image


src/pages/explore/ui/pair-card.tsx
```diff
@@ -7,76 +6,95 @@ import { AssetIcon } from '@penumbra-zone/ui/AssetIcon';
...
+        )}
       </div>
 
       <div className='h-10 flex flex-col items-end justify-center'>
         {loading ? (
           <ShimmeringBars />
         ) : (
           <>
-            <Text color='text.primary'>0.23</Text>
+            <Text color='text.primary'>{round({ value: summary.price, decimals: 6 })}</Text>
```
JasonMHasperhoven, line 95: I think the decimals here should come from quoteAssets exponent.

src/shared/api/server/summary/single.ts
```diff
@@ -2,10 +2,7 @@ import { NextRequest, NextResponse } from 'next/server';
 import { pindexer } from '@/shared/database';
```
JasonMHasperhoven, line 2: Not sure if I like the naming of these files,

How about we have routes like:
```
/summary/index.ts
/summary/[baseAsset]/[quoteAsset].ts
```

JasonMHasperhoven (approved):
LGTM, just a few minor points

hdevalence:
Some questions from looking at the screenshot (don't need to all be addressed in the PR necessarily but wanted to flag):

- it says USDY/USDC 2891T, this is because the denom exponents are not being handled
- the assets in the pairs are incorrectly ordered, the UI should always show UM/USDC rather than USDC/UM in the list. to handle this we will need the dex frontend to maintain a "ordering priority" / adjustment function that picks the "UI ordering" of a pair such that USDC always is asset 2, etc. (users should be able to navigate to the USDC/UM view with reversed ordering but in the list it should have the ordering we expect users want) 
- why is the "Liquidity" column measured in different tokens instead of in USDC? wouldn't it make more sense to show all in a common numeraire so the numbers are comparable? similarly, volume should be usd-denominated
- how does usdc/osmo have 0 24h volume but does have a 24h price change?

VanishMax:
@JasonMHasperhoven @hdevalence Thanks for your comments! I will address them all in the following PR

PR #171, opened 2024-12-01
Add an LP timeline view like the previous UI.

hdevalence:
This code is messy, but it should be merged ASAP, because not having the ability to inspect LP positions while doing development means there's no way to spot-check any other part of the UI, and that's not acceptable.

Requires https://github.com/penumbra-zone/penumbra/pull/4946 for the data backend.

Images:

Screenshot 2024-12-01 at 11 13 10 AM
Screenshot 2024-12-01 at 11 13 27 AM
Screenshot 2024-12-01 at 11 13 57 AM


hdevalence:
Added a breakdown of volume and fees by route, allowing LPs to see the performance of their positions and which routes are providing order flow

Screenshot 2024-12-01 at 2 56 34 PM
Screenshot 2024-12-01 at 3 02 00 PM
Screenshot 2024-12-01 at 2 56 50 PM

One of those positions has 11K executions and the page rendering is very bad. At some point this page should have pagination / loading in chunks for the executions but this should happen post-merge, getting some version of this up is a step towards closing the loop on making the DEX legible.

src/shared/database/index.ts
```diff
@@ -79,6 +100,159 @@ class Pindexer {
...
+    const results = await this.db
+      .selectFrom('dex_ex_position_executions')
+      .select([
+        'context_asset_start',
+        'context_asset_end',
+        sql`sum(delta_1 + lambda_1)`.as('volume1'),
+        sql`sum(delta_2 + lambda_2)`.as('volume2'),
+        sql`sum(fee_1)`.as('fees1'),
+        sql`sum(fee_2)`.as('fees2'),
+        sql`CAST(count(*) AS INTEGER)`.as('executionCount'),
```
VanishMax, line 240: suggestion: this would be better typed with the help of generics:

```ts
sql<number>`sum(delta_1 + lambda_1)`.as('volume1'),
```

meaning the returned `volume1` field would be of type `number`

grod220:
This was quite the colossal PR. The additional changes from me:
- Design-wise moved to a 2-column grid design to take advantage of more real estate
- Added loading skeletons
- Updated the search page logic + UI
- Changed backend to use
- Transitioned large one-file code into separate files (UI and backend)
- Assemble grpc ValueView's in the backend
- Update serialization helper
- Lots of type updates to appease Typescript
- Used more standard helpers available in packages

Given the size of this, opting to merge this versus get an additional review from the web team. Can iterate on this as we need.

grod220 (approved):

PR #173, opened 2024-12-02
feat: #111: explore page, trading pairs part 2

VanishMax:
References #111 

Changes from the part 1:
1. Large update of the SQL query to pindexer:
1.1. It now queries the summary table for pairs and joins it with the candles table
1.2. It now filters out the pairs where stablecoins are base assets, so only pairs of type "UM/USDC" are allowed, not "USDC/UM"
1.3. It aggregates latest 24 hours of candles per pair and returns it as an array
2. The aggregated candles often lose data on the way, so instead of 24 hourly candles frontend usually gets 1-15 candles. This PR introduces `adaptChart` function to fill the array of 24 data points for the chart, even having less values
3. `PreviewChart` component made with pure SVG as a `<polyline points={datapoints}>`.

Needs a follow up with part 3 for the pagination and search feature

image


src/pages/explore/ui/pair-card.tsx
```diff
@@ -58,8 +62,15 @@ export type PairCardProps =
     };
 
 export const PairCard = ({ loading, summary }: PairCardProps) => {
+  // TODO: change to correct dates
+  const today = subDays(new Date(), 3);
+  const yesterday = subDays(new Date(), 4);
```
VanishMax, line 67: Needs changes. I put this info because we have candles from 4 days ago right now but the data is completely missing from pindexer for this weekend

src/shared/api/server/summary/all.sql
VanishMax, line 1: i put the file here for simplicity of understanding of the `kysely` syntax for the explore page. Also, good for testing

src/pages/explore/ui/preview-chart/preview-adapter.test.ts
grod220, line 1: Thank you for these tests 🙏

src/shared/api/server/summary/all.sql
grod220, line 1: question: we don't need this right?

src/shared/api/server/summary/all.ts
```diff
@@ -59,17 +71,8 @@ export const GET = async (req: NextRequest): Promise<NextResponse<SummariesRespo
...
-        {
-          error: `durationWindow missing or invalid window. Options: ${durationWindows.join(', ')}`,
-        },
-        { status: 400 },
-      );
-    }
-
     const result = await getAllSummaries({
-      window,
+      window: window && isDurationWindow(window) ? window : '1d',
```
grod220, line 75: suggestion: if someone passes a window, I'd do validation+throw if `isDurationWindow()` returns a false instead of defaulting the query to `1d`. Or... we can just require the duration window passed by the client too.

src/shared/database/index.ts
grod220, line 1: holy sql!

grod220 (approved):

src/shared/api/server/summary/all.sql
VanishMax: I thought it could be good for testing but now I agree, removed this.

It is quite easy to get the SQL from kysely – just call `console.log(kysely.compile())` before `.execute()` – it would print the full SQL as a string

src/shared/api/server/summary/all.ts
VanishMax: agree, updated

PR #179, opened 2024-12-04
Take actions on positions

grod220:
Closes https://github.com/prax-wallet/prax/issues/237

This PR enables closing & withdrawing LP positions. 

Will open new issues for additional work:
- Proper price calculations
- UI updates to fix figma designs. Sizing & colors not quite right. Requires UI library changes.

https://github.com/user-attachments/assets/7baea7eb-35d9-49bc-9d06-5a63860b6956


src/shared/api/server/summary/types.ts
```diff
@@ -95,12 +95,12 @@ export class SummaryDataResponse {
...
   ): SummaryDataResponse {
     const directVolume = toValueView({
-      amount: summary.direct_volume_over_window,
+      amount: Math.floor(summary.direct_volume_over_window),
       metadata: quoteAsset,
     });
 
     const liquidity = toValueView({
-      amount: summary.liquidity,
+      amount: Math.floor(summary.liquidity),
```
grod220, line 103: A race to first review (cc https://github.com/penumbra-zone/dex-explorer/pull/178)

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+      const planReq = new TransactionPlannerRequest({
+        positionWithdraws: positions.map((positionId, i) => ({
+          positionId,
+          tradingPair: latestPositionData[i]?.data?.phi?.pair,
+          reserves: latestPositionData[i]?.data?.reserves,
+        })),
+        source: new AddressIndex({ account: connectionStore.subaccount }),
+      });
+
+      await planBuildBroadcast('positionWithdraw', planReq);
```
TalDerei, line 65: suggestion: `planBuildBroadcast` may fail, can we add an [errorToast](https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/state/ibc-out.ts#L111)? The subsequent `updatePositionsInCache` could also fail.

src/pages/trade/ui/positions.tsx
```diff
@@ -39,7 +56,132 @@ const ErrorNotice = ({ error }: { error: unknown }) => {
...
+      valueView={kind === 'tradeAmount' ? o.tradeAmount : o.effectivePrice}
+      key={i}
+    />
+  ));
+};
+
+const MAX_ACTION_COUNT = 15;
+
+const HeaderActionButton = observer(() => {
+  const { data } = usePositions();
```
TalDerei, line 146: should the `useQuery` have a [retry](https://tanstack.com/query/latest/docs/framework/react/guides/query-retries)?

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+class PositionsStore {
+  public loading = false;
+
+  constructor() {
+    makeAutoObservable(this);
+  }
+
+  setLoading(state: boolean) {
+    this.loading = state;
+  }
```
TalDerei, line 22: comment: I've been reading about https://medium.com/@brojd/automate-loading-statuses-in-mobx-d6965ad738d9 and https://mobx.js.org/actions.html#using-flow-instead-of-async--await- for automated loading and using `flow` annotation instead of async / await for mobx state management. this might be something interesting to look into the future?

TalDerei (approved):
some minor comments

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+
+      await planBuildBroadcast('positionClose', planReq);
+      await this.updatePositionsInCache(positions);
+    } catch (e) {
+      openToast({
+        type: 'error',
+        message: 'Error with withdraw action',
+        description: String(e),
+      });
```
TalDerei, line 40: should we port over the [errorToast](https://github.com/prax-wallet/prax/blob/main/packages/ui/lib/toast/presets.ts#L10) utility from prax?

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // After a successful action, update the position state in the cache
+  private async updatePositionInCache(positionId: PositionId) {
+    const { data } = await penumbra.service(DexService).liquidityPositionById({ positionId });
+    if (data) {
+      const newPositionData = await getPositionData(positionId, data);
+      queryClient.setQueryData<PositionData[]>(['positions'], oldData => {
+        if (!oldData) {
+          return undefined;
```
TalDerei, line 90: should this always return `[newPositionData]`?

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+import { planBuildBroadcast } from '@/pages/trade/ui/order-form/helpers.tsx';
+import { connectionStore } from '@/shared/model/connection';
+import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { queryClient } from '@/shared/const/queryClient.ts';
+import { getPositionData, PositionData } from '@/pages/trade/api/positions.ts';
+import { penumbra } from '@/shared/const/penumbra.ts';
+import { DexService } from '@penumbra-zone/protobuf';
+import { openToast } from '@penumbra-zone/ui/Toast';
+
+class PositionsStore {
```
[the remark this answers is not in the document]
grod220: ah, will move 👍

src/pages/trade/store/positions.ts
grod220: Interesting! No idea, but let's continue to explore good async patterns like this 👍

src/pages/trade/store/positions.ts
grod220: Probably worth a group discussion how people think about the toast api

src/pages/trade/store/positions.ts
```diff
@@ -0,0 +1,99 @@
...
+      this.setLoading(false);
+    }
+  };
+
+  private async updatePositionsInCache(positions: PositionId[]): Promise<void> {
+    const promises = positions.map(id => this.updatePositionInCache(id));
+    await Promise.all(promises);
+  }
+
+  // After a successful action, update the position state in the cache
```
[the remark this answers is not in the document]
grod220: Good question! For most things, we should use `useRefetchOnNewBlock()`. However, for positions, the position state isn't stored locally (context: https://github.com/penumbra-zone/penumbra/pull/4837) and must be fetched freshly from the node. That likely means it is too expensive to refetch all of our position states every block (or maybe not?).

This `updatePositionInCache` function exists to specifically re-fetch the position we know changed and update only that in the cache as a means of not having to re-fetch all of your positions states again.

src/pages/trade/store/positions.ts
grod220: Aha, this is a bit tricky. In theory, a component should not be trying to use this if they have not first fetched for positions already.

If we were to return `[newPositionData]`, the next component that would use `usePositions()` would get an incorrect cache (the other positions are not fetched).

Think the best way to handle this is to throw to inform the dev they've made a mistake instead of ignoring the misuse.

src/pages/trade/ui/history-tabs.tsx
```diff
@@ -24,22 +22,18 @@ export const HistoryTabs = () => {
...
+            onChange={value => setTab(value as PositionsTabsType)}
+            options={[{ value: PositionsTabsType.MY_POSITIONS, label: 'My Positions' }]}
           />
         </Density>
 
         <label className='flex gap-2 h-[42px] items-center py-2 text-text-secondary cursor-pointer'>
-          <Text small>Show all</Text>
-          <Toggle label='Show all' value={showAll} onChange={setShowAll} />
+          <Text small>Show Inactive</Text>
+          <Toggle label='Show Inactive' value={showInactive} onChange={setshowInactive} />
```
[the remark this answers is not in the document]
grod220: Plenty of UI inconsistencies in this PR 😬

src/pages/trade/ui/positions.tsx
```diff
@@ -39,7 +56,132 @@ const ErrorNotice = ({ error }: { error: unknown }) => {
...
+  }
+  return p.orders.map((o, i) => (
+    <ValueViewComponent
+      valueView={kind === 'tradeAmount' ? o.tradeAmount : o.effectivePrice}
+      key={i}
+    />
+  ));
+};
+
+const MAX_ACTION_COUNT = 15;
```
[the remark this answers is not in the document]
grod220: Yes. We had a `Close All` button in minifront for auctions and it proved to be too expensive for anything above ~15 actions or so. This is an attempt to learn from those mistakes.

src/shared/api/server/summary/types.ts
```diff
@@ -95,12 +95,12 @@ export class SummaryDataResponse {
     candleTimes?: Date[],
   ): SummaryDataResponse {
     const directVolume = toValueView({
-      amount: summary.direct_volume_over_window,
+      amount: Math.floor(summary.direct_volume_over_window),
```
JasonMHasperhoven, line 98: Why to use floor over round here btw?

src/pages/trade/ui/positions.tsx
grod220: The default actually retries like 5 times I think. We should figure out a way to change this default to retry less because the loading states are too long in a failure case imo. Going to change this to `retry: 1` for now.

src/pages/trade/store/positions.ts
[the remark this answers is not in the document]
TalDerei: makes sense, thanks for the clarification 👍

PR #181, opened 2024-12-04
Limit Orders

JasonMHasperhoven:


hdevalence:
Screenshot 2024-12-09 at 6 43 41 PM

Running locally I get summary issues -- I think this was already fixed in `main`, do we think it would disappear on rebase?

cronokirby:
> Screenshot 2024-12-09 at 6 43 41 PM
> 
> Running locally I get summary issues -- I think this was already fixed in `main`, do we think it would disappear on rebase?

I got this error when I forgot to set the environment variable to point at a node

src/shared/math/position.ts
```diff
@@ -62,24 +63,55 @@ const priceToPQ = (
...
+  const p = pnum(
+    BigInt(
+      BigNumber((qExponentUnits * scale).toString())
+        .times(BigNumber(price))
+        .shiftedBy(-(qExponent - pExponent))
+        .toFixed(0),
+    ),
+    qExponent,
+  ).toAmount();
+
```
erwanor, line 85: Do we need `scale` at all?

```typescript
// Pseudo code
const p_1 = round(quoteAssetExponent * price);
const p_2 = baseAssetExponent;
```

erwanor: I don't understand the exponent shift.

src/shared/math/position.ts
```diff
@@ -0,0 +1,218 @@
...
+  return new Position({
+    phi: {
+      component: {
+        fee: plan.feeBps,
+        // p: pnum(p).toAmount(),
+        // q: pnum(q).toAmount(),
+        p,
+        q,
+      },
+      pair: new TradingPair({
```
erwanor, line 116: Does this perform any validation or re-ordering?

Asset IDs are essentially large numbers, and a canonical trading pair for two assets $A, B$ is going to be based on ordering, suppose $B < A$
- asset 1 is $B$
- asset 2 is $A$

and vice-versa

src/shared/math/position.ts
```diff
@@ -0,0 +1,218 @@
...
+ * as an escape hatch in case any of those use cases aren't sufficient.
+ */
+export const planToPosition = (plan: PositionPlan): Position => {
+  console.log('TCL: plan', plan);
+  // const { p, q } = priceToPQ(plan.price, plan.baseAsset.exponent, plan.quoteAsset.exponent);
+  const { p, q } = priceToPQ2(plan.price, plan.baseAsset.exponent, plan.quoteAsset.exponent);
+
+  // const r1 = pnum(plan.baseReserves, plan.baseAsset.exponent).toAmount();
+  // const r2 = pnum(plan.quoteReserves, plan.quoteAsset.exponent).toAmount();
+  const r1 = pnum(plan.baseReserves).toAmount();
```
erwanor, line 104: I think we need to make sure that base is asset_1 and quote is asset_2 and flip everything if it's not.

Ideally, there's a `Position` constructor that does this on your behalf so that you don't have to worry about it.

hdevalence:
Some things I tested but couldn't get working:

1. Go to https://localhost/trade/USDY/USDC, try to create a "buy" limit order to buy 1 USDY @ 1:1 using 1 USDC. this fails with a negative exponent error

2. Same page, opposite direction, try to create a "sell" limit order to sell 1 USDC @ 1:1. this fails first with a missing source field (was undefined, pushed a commit that works around though i'm not sure it's the right approach), then I got an error that I didn't have any USDY. Reserves don't look correct to me, but I'm not sure why?

USDY/USDC limit orders are a good test case because they exercise different denoms and also unlock a market-based solution to the USDC transfer issue.

Screenshot 2024-12-10 at 7 22 24 PM
Screenshot 2024-12-10 at 7 22 46 PM
Screenshot 2024-12-10 at 7 23 07 PM

cronokirby:
I refactored the stores to all have a much more consistent set of transitions and a reduced set of truth, and everything now seems to be working correctly.

I clobbered some work in the process, which is backed up in https://github.com/penumbra-zone/dex-explorer/tree/limitorders-archive just in case.

There were a handful of bugs not even reported in slack across the various forms, but now it should all work.

I tested it buy working on the testnet, using TestUSD and UM, which have 1e18 and 1e6 as the exponent

**Range Liquidity:**
image
image
The amount of um and test usd being consumed is right on the website, but also in the extension. Also, the base unit price is 1.7e12 in the first screenshot, which is accurate, given the exponent difference.

**Limit Order**
_Buy_
image
image
_Sell_
image
image

Some testing on mainnet would be appreciated

grod220 (approved):
Given the number of devs that have contributed to this PR, think we should feel confident to ship and iterate as needed in follow ups 👍

PR #185, opened 2024-12-05
feat: serialize

VanishMax:
This PR implements helper functions to deal with Protobuf serialization:
- `serialize`
- `deserialize`
- `innerFetch`

Also it fixes the `stats` request that wasn't working when published to the servers. The problem with RSC (react server components) is that we actually can't request fresh data – the RSC requests happen in **build-time**, which is undesirable.

Type-safe handlers help reducing boilerplate and simplify reading the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "dev": "next dev",
     "dev:local": "WEBPACK_WATCH=true next dev",
     "start": "next start",
+    "preview": "next build && cp -r public .next/standalone/ && cp -r .next/static .next/standalone/.next/ && cp -r ca-cert.pem .next/standalone/ && node .next/standalone/server.js",
     "format": "prettier --write .",
     "lint": "eslint src",
     "lint:fix": "eslint src --fix",
--- a/src/pages/explore/api/use-stats.ts
+++ b/src/pages/explore/api/use-stats.ts
@@ -1,33 +1,14 @@
 'use client';
 
 import { useQuery } from '@tanstack/react-query';
-import type { StatsDataJSON, StatsData, StatsResponseJSON } from '@/shared/api/server/stats';
-import { ValueView } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import type { StatsData } from '@/shared/api/server/stats';
+import { innerFetch } from '@/shared/utils/inner-fetch';
 
 export const useStats = () => {
-  const { data, ...rest } = useQuery<StatsDataJSON>({
+  return useQuery<StatsData>({
     queryKey: ['stats'],
     queryFn: async () => {
-      const baseUrl = '/api/stats';
-      const fetchRes = await fetch(baseUrl);
-      const jsonRes = (await fetchRes.json()) as StatsResponseJSON;
-      if ('error' in jsonRes) {
-        throw new Error(jsonRes.error);
-      }
-
-      return jsonRes;
+      return innerFetch<StatsData>('/api/stats');
     },
   });
-
-  const stats = data && {
-    ...data,
-    largestPairLiquidity: data.largestPairLiquidity ? ValueView.fromJson(data.largestPairLiquidity) : undefined,
-    liquidity: ValueView.fromJson(data.liquidity),
-    directVolume: ValueView.fromJson(data.directVolume),
-  } satisfies StatsData;
-
-  return {
-    ...rest,
-    data: stats,
-  };
 };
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "dev": "next dev",
     "dev:local": "WEBPACK_WATCH=true next dev",
     "start": "next start",
+    "preview": "next build && cp -r public .next/standalone/ && cp -r .next/static .next/standalone/.next/ && cp -r ca-cert.pem .next/standalone/ && node .next/standalone/server.js",
```
VanishMax, line 17: This method starts the standalone nexts app. It's needed since `next start` is made for static-generated apps

src/pages/explore/api/use-stats.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { useQuery } from '@tanstack/react-query';
+import type { StatsData } from '@/shared/api/server/stats';
+import { innerFetch } from '@/shared/utils/inner-fetch';
+
+export const useStats = () => {
+  return useQuery<StatsData>({
+    queryKey: ['stats'],
+    queryFn: async () => {
+      return innerFetch<StatsData>('/api/stats');
```
VanishMax, line 11: the usage of `innerFetch` function

src/shared/utils/serializer.ts
```diff
@@ -0,0 +1,75 @@
...
+  value: JsonValue;
+}
+
+// Type utility to serialize message fields to strings
+export type Serialized<T> =
+  T extends Message<any>
+    ? SerializedProto
+    : T extends object
+      ? { [KEY in keyof T]: Serialized<T[KEY]> }
+      : T;
```
VanishMax, line 17: usage:

```ts
// correct – no need to wrap this in `serialize` function
const errorObj = { error: '' } satisfies Serialized;

// incorrect - needs a wrapper
const protoObj = { value: new ValueView() } satisfies Serialized;

// correct - has a wrapper
const protoObj = serialize({ value: new ValueView() }) satisfies Serialized;
```

src/shared/utils/serializer.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  return obj as Serialized<VAL>;
+};
+
+const ProtosByType = {
+  'penumbra.core.asset.v1.ValueView': ValueView,
+  'penumbra.core.asset.v1.Metadata': Metadata,
+} as const;
```
VanishMax, line 42: this should be extended with time

package.json
grod220: question: this replicates running a production build?

src/shared/utils/inner-fetch.ts
```diff
@@ -0,0 +1,25 @@
+import { deserialize, Serialized } from './serializer';
+
+/**
+ * A wrapper around `fetch` to request data from local endpoints. Features:
+ * 1. Composes the URL search params correctly, only provide an object
+ * 2. Throws if the response contains an error object
+ * 3. Deserializes the response object, if it contains protobuf messages
+ */
+export const innerFetch = async <RES extends object>(
```
grod220, line 9: suggestion: `inner` feels a bit like a private field. I wonder if `apiFetch` or `fetchWithSerialization` or `structuredFetch` may be better.

src/pages/explore/api/use-stats.ts
```diff
@@ -0,0 +1,14 @@
+'use client';
+
+import { useQuery } from '@tanstack/react-query';
+import type { StatsData } from '@/shared/api/server/stats';
+import { innerFetch } from '@/shared/utils/inner-fetch';
+
+export const useStats = () => {
+  return useQuery<StatsData>({
+    queryKey: ['stats'],
```
grod220, line 9: question: does this need a duration window?

grod220 (approved):
Nice job. Think this will help in the immediate term. However, after we transition to using proper grpc api infra, this will be phased out.

src/shared/utils/serializer.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const deserializeProto = (value: SerializedProto): Message<any> => {
```
TalDerei, line 44: :( pitfall city

package.json
VanishMax: yes, the same script as in the `Containerfile` and as explained here: https://nextjs.org/docs/pages/api-reference/next-config-js/output#automatically-copying-traced-files

src/pages/explore/api/use-stats.ts
VanishMax: it is hardcoded in sql, don't think we'll need dynamic window for this

src/shared/utils/inner-fetch.ts
VanishMax: thanks, `apiFetch` feels like a better name

PR #187, opened 2024-12-05
fix: routebook visual updates

### LINKED ISSUE #146: Route book updates
It's currently too tight:

Screenshot 2024-11-21 at 5 43 06 PM

Figma designs (ignore other font for now)
Screenshot 2024-11-21 at 5 43 02 PM

Also should be using the skeleton from the summary component and delete the one currently in use.

vacekj:
fixes #146 
- Adjusted padding in routebook
- Replace custom skeleton with the reusable skeleton
- Adjusted formatting of numbers to conform with Figma and fit on smaller screens, including the padding.

### DIFF
--- a/src/pages/trade/ui/route-book.tsx
+++ b/src/pages/trade/ui/route-book.tsx
@@ -61,12 +61,8 @@ const RouteBookData = observer(({ bookData }: { bookData?: RouteBookResponse }) 
     <div className='w-full'>
       {/* Header */}
       <div className='grid grid-cols-[1fr_1fr_1fr_1fr] h-[33px] mt-1 text-xs text-gray-400 px-4'>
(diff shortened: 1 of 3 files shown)


VanishMax (changes requested):
Ran the app, found some visual issues:
1. Column "total" misalignment – the number should be aligned with the right side of a cell
2. Columns go out of the screen, "route" column doesn't fit.
3. Hops amount takes multiple lines
4. It seems like the background doesn't fully fit the hovered row (last screen, in the bottom of it there's a black line that shouldn't be there)

image

image

vacekj:
@VanishMax apologies - it seems I pushed a botched up version, working on a fix now.

VanishMax (approved):
Nice! Grid looks like a better solution here.

Needs linting and it is good to go

PR #189, opened 2024-12-06
fix: remove redirect loader

VanishMax:
Previously, we were redirecting users from "/" and "/trade" pages to "/trade/UM/USDC" using client-side components. It made the app firstly load empty screen, wait for the whole bunch of JavaScript to process, and only then send a fetch request for a default asset pair. It annoyed me, so I changed the redirect to be server-side using Next middleware. Feels like a lightning speed!

As a plus, added the `not-found` page.

The next video shows before and after.

https://github.com/user-attachments/assets/6f150cf7-2664-42d8-8cf5-f7af78327d88

Not found page

image


grod220 (approved):
what the heck that is so much faster! 🏎️ 🔥

PR #193, opened 2024-12-09
feat: add volume data

### LINKED ISSUE #153: Add volume profile to chart
- Docs: https://tradingview.github.io/lightweight-charts/tutorials/how_to/price-and-volume
- jsfiddle: https://jsfiddle.net/TradingView/cnbamtuh/

vacekj:
fixes #153

### DIFF
--- a/src/pages/trade/ui/chart.tsx
+++ b/src/pages/trade/ui/chart.tsx
@@ -1,11 +1,12 @@
 import cn from 'clsx';
 import { useEffect, useRef, useState } from 'react';
-import { createChart, IChartApi, OhlcData } from 'lightweight-charts';
+import { createChart, IChartApi } from 'lightweight-charts';
 import { theme } from '@penumbra-zone/ui/theme';
 import { Text } from '@penumbra-zone/ui/Text';
 import { useCandles } from '../api/candles';
 import { observer } from 'mobx-react-lite';
 import { DurationWindow, durationWindows } from '@/shared/utils/duration.ts';
+import { CandleWithVolume } from '@/shared/api/server/candles/utils';
 
 const ChartLoadingState = () => {
   return (
@@ -119,10 +120,11 @@ const ChartLoadingState = () => {
   );
 };
 
-const ChartData = observer(({ candles }: { candles: OhlcData[] }) => {
+const ChartData = observer(({ candles }: { candles: CandleWithVolume[] }) => {
   const chartElRef = useRef<HTMLDivElement>(null);
   const chartRef = useRef<IChartApi>();
   const seriesRef = useRef<ReturnType<IChartApi['addCandlestickSeries']>>();
+  const volumeSeriesRef = useRef<ReturnType<IChartApi['addHistogramSeries']>>();
 
   // Initialize the chart once when the component mounts
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/shared/api/server/candles/utils.ts
```diff
@@ -2,22 +2,30 @@ import { OhlcData, UTCTimestamp } from 'lightweight-charts';
...
-    open: c.open,
-    time: (c.start_time.getTime() / 1000) as UTCTimestamp,
+    ohlc: {
+      close: c.close,
+      high: c.high,
+      low: c.low,
+      open: c.open,
+      time: (c.start_time.getTime() / 1000) as UTCTimestamp,
+    },
+    volume: c.direct_volume + c.swap_volume,
```
grod220, line 19: question: can we validate with someone this is the correct way to get volume? For example, isn't swap volume the multi-hop route (?), and if so, should that count?

grod220 (commented):
Nice job!

Two design points of feedback:
- Could we shorten the height a bit? Feels a bit too tall. Looking at [something like this](https://www.lcx.com/wp-content/uploads/FireShot-Capture-116-Trades-terminal-staging.lcx_.com_.png) as reference.
- Could we dim the red/green color so it doesn't look like it's competing with the price candles? 

It looks like there is another label on the right, can we get rid of this?
Screenshot 2024-12-09 at 5 21 45 PM

src/pages/trade/ui/chart.tsx
```diff
@@ -119,10 +120,11 @@ const ChartLoadingState = () => {
   );
 };
 
-const ChartData = observer(({ candles }: { candles: OhlcData[] }) => {
+const ChartData = observer(({ candles }: { candles: CandleWithVolume[] }) => {
   const chartElRef = useRef<HTMLDivElement>(null);
   const chartRef = useRef<IChartApi>();
   const seriesRef = useRef<ReturnType<IChartApi['addCandlestickSeries']>>();
+  const volumeSeriesRef = useRef<ReturnType<IChartApi['addHistogramSeries']>>();
```
grod220, line 127: question: curious, do they ever overlap? What would happen?

vacekj: I added the scaling code so they don't overlap, however they can overlap if there is extremely high volume and a big drop in price. I haven't encountered this so far.

src/shared/api/server/candles/utils.ts
vacekj: Looks like this is correct.

hdevalence:
Haven't tested the branch, but looking at the screenshot, the chart looks like the UM/USDC chart but the volume is listed as 19.5B, which doesn't seem right?

src/pages/trade/ui/summary.tsx
```diff
@@ -95,7 +95,7 @@ export const Summary = () => {
       </SummaryCard>
       <SummaryCard title='24h Volume' loading={isLoading}>
         {data && 'directVolume' in data ? (
-          <Density slim>
+          <Density compact>
```
grod220, line 98: issue: this should be `slim` as the token icon gets too big if not, can you revert back?

grod220 (approved):
Nice job!

src/pages/trade/ui/summary.tsx
vacekj: This was causing a lint error for me, might've been on my side. Will fix

Issue #155, opened 2024-11-22
Relevant accounts displayed only


grod220:
At the moment we are hardcoding the sub-account numbers: https://github.com/penumbra-zone/dex-explorer/blob/455e6b2b9dec3e2305d358a61a3ea669f2be3dbf/src/widgets/header/api/subaccounts.ts#L10.

We should default to `Main Account` and also call the balance rpc to the extension and use that to only display sub-accounts where we have balances. 

<img width="346" alt="Screenshot 2024-11-22 at 1 42 06 PM" src="https://github.com/user-attachments/assets/379e89b8-65cf-4bc3-80d5-a4a330341d30">

PR #202, opened 2024-12-11
ui: display relevant accounts

TalDerei:
references https://github.com/penumbra-zone/dex-explorer/issues/155

`useSubaccounts` react-query hook is tracked by mobx observer HOC, and queries account balances from the view service to display the relevant accounts. 

the `Address View` V2 UI component (https://github.com/penumbra-zone/web/pull/1941) was implemented towards realizing this. 

-------------------------------

Screenshot 2024-12-10 at 11 25 55 PM

----------------------------

https://github.com/user-attachments/assets/4d90365e-6a55-4279-b9d7-2cc85a8b0a13


src/widgets/header/api/subaccounts.ts
```diff
@@ -30,8 +45,13 @@ const fetchQuery = async (): Promise<AddressView[]> => {
...
 
 export const useSubaccounts = () => {
+  // Query account balances from view service
+  const { data: balances } = useBalances();
+
   return useQuery({
     queryKey: ['view-service-accounts'],
-    queryFn: fetchQuery,
+    queryFn: () => {
+      return fetchQuery(balances!);
```
TalDerei, line 54: can't seem to resolve this non-null assertion linting error. Using a nullish coalescing operator (`balances ?? []`) here will cause the widget to break and only the main account in the UI.

grod220: How about this?

```ts
  return useQuery({
    queryKey: ['view-service-accounts', balances],
    queryFn: () => {
      if (!balances) {
        return []
      }
      return fetchQuery(balances);
    },
  });
```

We should make `balances` a query key btw, that way if a new balances object comes back, we refetch.

package.json
```diff
@@ -36,14 +36,14 @@
...
-    "@penumbra-zone/client": "^21.0.0",
+    "@penumbra-zone/client": "^21.1.0",
     "@penumbra-zone/crypto-web": "^30.0.0",
     "@penumbra-zone/getters": "^20.0.0",
     "@penumbra-zone/perspective": "^40.0.0",
     "@penumbra-zone/protobuf": "^6.3.0",
     "@penumbra-zone/transport-dom": "^7.5.0",
     "@penumbra-zone/types": "^26.3.0",
-    "@penumbra-zone/ui": "^13.4.2",
+    "@penumbra-zone/ui": "^13.5.0",
```
grod220, line 46: Looks like we have some density issues:

Screenshot 2024-12-11 at 2 28 57 PM

grod220 (commented):
It looks like there is a flicker when loading. Could we have a spinner in the button to indicate that status? 

Screenshot 2024-12-11 at 2 29 20 PM

package.json
TalDerei: density updates from the UI package bump are addressed in https://github.com/penumbra-zone/dex-explorer/pull/204 cc @VanishMax

src/widgets/header/api/subaccounts.ts
TalDerei: making balance a query key resolves this

src/widgets/header/ui/subaccount-selector.tsx
```diff
@@ -23,7 +24,11 @@ export const SubaccountSelector = observer(() => {
       <DropdownMenu>
         <DropdownMenu.Trigger>
           <Button>
-            <AddressViewComponent copyable={false} addressView={valueAddress} />
+            {isLoading ? (
+              <SpinnerIcon className='animate-spin' />
```
grod220, line 28: Suggestion: When it loads, the `main acccount` state of the button is ~170px wide. Think we should make the button width in the loading state the same width so there is less screen jank as it transitions from loading to loaded.

grod220 (approved):
Nice job!

src/widgets/header/ui/subaccount-selector.tsx
TalDerei: noting that a width of `~142px` fits perfectly on screen reloads, but I adjusted it to `170px` to match the button width.

PR #207, opened 2024-12-12
Order form bug fix

grod220:
Currently throwing when wallet attempts to reconnect:

https://github.com/user-attachments/assets/8590e3b0-3a7a-465c-be10-7e35cedf6492


grod220:
Given the disturbance to dev for folks, going to force merge this

PR #208, opened 2024-12-12
Don't calculate in constructor

grod220:
Currently the cause of an error toast on load:

Screenshot 2024-12-12 at 9 48 17 PM

### DIFF
--- a/src/pages/trade/ui/order-form/store/index.ts
+++ b/src/pages/trade/ui/order-form/store/index.ts
@@ -57,7 +57,6 @@ class OrderFormStore {
     makeAutoObservable(this);
 
     void this.calculateGasFee();
-    void this.calculateExchangeRate();
   }
 
   setType = (type: FormType): void => {


src/pages/trade/ui/order-form/store/index.ts
```diff
@@ -57,7 +57,6 @@ class OrderFormStore {
     makeAutoObservable(this);
 
     void this.calculateGasFee();
-    void this.calculateExchangeRate();
```
grod220, line 60: @JasonMHasperhoven can you validate this is ok? On load, there is no assetIds set so it is throwing. Upon the assetIds loading in the store, it re-renders and runs the calculations later anyway.

TalDerei: on that note, why are we even running `void this.calculateGasFee()` in the constructor? It similarly would be undefined

grod220: Yeah, we kinda have a race condition with these. I think we can probably delete both.

src/pages/trade/ui/order-form/store/index.ts
```diff
@@ -57,7 +57,6 @@ class OrderFormStore {
     makeAutoObservable(this);
 
     void this.calculateGasFee();
```
TalDerei, line 59: suggestion: remove from constructor

TalDerei (approved):

JasonMHasperhoven:
This is completely refactored in https://github.com/penumbra-zone/dex-explorer/pull/181, so I think we should close this.

PR #215, opened 2024-12-13
Over-fetching hook

grod220:
The `useRefetchOnNewBlock()` is overfetching as there isn't a cache for the many re-render calls happening. 

Before/after fix:

https://github.com/user-attachments/assets/29b4b3c4-157a-4f41-ab47-546598a16ca9


TalDerei (approved):
really great improvement! 

so basically refetch is only called once per block height for a given query key

PR #218, opened 2024-12-14
fix: reload on wallet disconnect

### LINKED ISSUE #157: Post-disconnect asset list persisting
Steps to reproduce:
1. Connect wallet that has a balance
2. Click asset selector and note token balances
3. Disconnect wallet
4. Click asset selector tab and the previous connected wallet's balances persist (refreshing fixes it)

Very likely a mobx store not re-rendering after the connection state changes

Example after disconnecting:
Screenshot 2024-11-22 at 4 42 57 PM

vacekj:
fixes #157 by reloading on wallet disconnect

### DIFF
--- a/src/shared/model/connection/index.ts
+++ b/src/shared/model/connection/index.ts
@@ -83,9 +83,10 @@ class ConnectionStateStore {
 
     try {
       await penumbra.disconnect();
-      window.location.reload();
     } catch (error) {
       console.error(error);
+    } finally {
+      window.location.reload();
     }
   }
 
--- a/src/shared/model/connection/index.ts
+++ b/src/shared/model/connection/index.ts
@@ -83,6 +83,7 @@ class ConnectionStateStore {
 
     try {
       await penumbra.disconnect();
+      window.location.reload();
     } catch (error) {
       console.error(error);
     }


src/shared/model/connection/index.ts
```diff
@@ -83,6 +83,7 @@ class ConnectionStateStore {
 
     try {
       await penumbra.disconnect();
+      window.location.reload();
```
TalDerei, line 86: suggestion: we want to ensure the page reload is handled in the `finally` block, independent of whether the disconnection succeeds or fails.

```ts
    try {
      await penumbra.disconnect();
    } catch (error) {
      console.error("Error during disconnect:", error);
    } finally {
      window.location.reload();
    }
```

consider for instance when the extension's connection context becomes invalidated, resulting in an unsuccessful disconnection.

https://github.com/user-attachments/assets/7f846f0e-2df3-467d-bfe2-ae6444c20812

TalDerei (changes requested):
minor nit

src/shared/model/connection/index.ts
vacekj: Good point, fixed.

TalDerei (approved):

Issue #188, opened 2024-12-05
Data issues with explore page


hdevalence:
Highlighted some issues with the explore page:

<img width="959" alt="Screenshot 2024-12-05 at 1 06 01 PM" src="https://github.com/user-attachments/assets/14aeda86-c026-47db-8c5e-b77c3470e1ef">

Big picture here is: we need to be treating all the values of liquidity, volume, etc, as numbers in the base unit of whatever asset, then using price info to convert to amounts of the common numeraire (let's use USDC for now) and showing that.

PR #222, opened 2024-12-17
Display recent executions

grod220:
Populating view for:

Screenshot 2024-12-17 at 12 43 08 PM

Note, route data is not available in pindexer.

PR #204, opened 2024-12-11
feat: #163: pair selector

VanishMax:
Closes #163 

Rewamped pair selector with:
- Starring feature
- Displaying pairs from backend with volume represented in USDC
- Searching pairs by symbol and by name
- Saving recent searches
- Loading and error states

https://github.com/user-attachments/assets/df7054be-2d37-4581-a36f-be5152b4c21c

PR #223, opened 2024-12-18
explore page: use common numeraire

TalDerei:
references https://github.com/penumbra-zone/dex-explorer/issues/188

- migrate `useSummary` useQuery into `/trade/api`,
- summaries: Implements a left join of the summary with the USDC prices table in the SQL query instead of making multiple internal API server requests to `/api/summary`, and denominates the trading pair liquidity and volume in USDC,
- stats: Implements a left join of the summary with the USDC prices table in the SQL query, and denominates the stats in USDC,
- fixes internal server error in `api/stats`,
- filters out pairs with zero liquidity and trading volume,
- utility: split USDC price calculations into separate utility method


TalDerei:
It seems that the `/api/stats` and `/api/summaries` endpoints from pindexer may be returning incorrect information, impacting metrics like "total trading volume (24h)" and the largest "trading pair (24h volume)" for the USDC/USDY pair. Previous [discussions](https://github.com/penumbra-zone/dex-explorer/pull/133#discussion_r1856531144) suggest that this might not just be a denom-related issue.

src/pages/trade/model/useSummary.ts
```diff
@@ -24,3 +24,24 @@ export const useSummary = (window: DurationWindow) => {
...
 };
+
+// Fetch USDC-denominated prices for each quote asset to enable cross-pair comparisons.
+// Uses 'useQueries' hook for handling multiple dynamic numbers of queries
+export const useMultipleSummaries = (quoteAssets: string[]) => {
+  const queries = useQueries<Array<{ data: SummaryData | NoSummaryData | undefined }>>({
+    queries: quoteAssets.map(asset => ({
+      queryKey: ['multipleSummaries', asset, 'USDC'] as const,
+      queryFn: async () => {
+        return apiFetch<SummaryData | NoSummaryData>('/api/summary', {
```
TalDerei, line 35: It seems that the` /api/stats` and `/api/summaries` endpoints from pindexer may be returning incorrect information, impacting metrics like "total trading volume (24h)" and the largest "trading pair (24h volume)" for the USDC/USDY pair. 

previous [discussions](https://github.com/penumbra-zone/dex-explorer/pull/133#discussion_r1856531144) suggest that this may not simply be related to the denom exponents not being handled.

src/pages/trade/model/useSummary.ts
```diff
@@ -24,3 +24,24 @@ export const useSummary = (window: DurationWindow) => {
...
+          durationWindow: '1d' as DurationWindow,
+          baseAsset: asset,
+          quoteAsset: 'USDC',
+        });
+      },
+    })),
+  });
+
+  // TODO: add block-based fetching
```
TalDerei, line 44: we can't simply call `useRefetchOnNewBlock('summary', queries)` since queries is a `UseQueryResult[]`

src/shared/utils/price-conversion.ts
```diff
@@ -22,3 +23,44 @@ export const calculateDisplayPrice = (
...
+    });
+  }
+
+  return quoteAssetsList;
+};
+
+/**
+ * Calculates USDC-normalized amounts for trading pair values (volume or liquidity)
+ */
+export const calculateScaledAmount = (
```
TalDerei, line 47: shared price conversion utility

src/pages/explore/ui/pairs.tsx
```diff
@@ -51,8 +54,49 @@ export const ExplorePairs = () => {
 
   const [search, setSearch] = useState('');
 
+  // Fetch trading pair summaries.
   const { data, isLoading, isRefetching, isFetchingNextPage, fetchNextPage } = useSummaries(search);
```
TalDerei, line 58: Is it possible that the pairs displayed on the explore page occasionally flip? Compare `SHITMOS / UM` versus `UM / SHITMOS` for instance

------------------------

'Screenshot 2024-12-17 at 12 28 19 PM

Screenshot 2024-12-17 at 8 40 06 PM

src/pages/explore/api/use-summaries.ts
```diff
@@ -5,7 +5,8 @@ import { SummaryData } from '@/shared/api/server/summary/types';
 import { DurationWindow } from '@/shared/utils/duration';
 import { apiFetch } from '@/shared/utils/api-fetch';
 
-const BASE_LIMIT = 15;
+/// The base limit will need to be increased as more trading pairs are added to the explore page.
+const BASE_LIMIT = 20;
```
TalDerei, line 9: caps the number of pairs returned from calling `useSummaries();`

src/pages/explore/ui/pairs.tsx
```diff
@@ -51,8 +54,49 @@ export const ExplorePairs = () => {
...
+  });
+
+  // Update volume and liquidity amounts with their respective USDC-normalized values,
+  // skipping USDC pairs.
+  if (translator) {
+    translator.forEach(page => {
+      page.forEach((item, itemIndex) => {
+        if (quoteAssetsList[itemIndex] !== 'USDC') {
+          item.directVolume.valueView.value!.amount!.lo = volumePairs[itemIndex]!;
+          item.liquidity.valueView.value!.amount!.lo = liquidityPairs[itemIndex]!;
```
TalDerei, line 94: this assignment is what updates the volume and liquidity values to be displayed on the explore page.I think the page should complete the price conversion and asset loading before rendering. Otherwise, the page may load first, and the amounts will update afterward. That's not currently the case, and should be updated to reflect that accordingly.

https://github.com/user-attachments/assets/6397f5ff-079c-40b4-a1ed-09f428edbcdf

VanishMax (commented):
question: is it possible to implement USDC price calculation using SQL? 

As I understood the code correctly, it firstly calls `useSummaries`, waits for it to fetch, and then makes multiple requests to `/api/summary` with USDC as a quote asset. In my tests, these sequential requests run from 1s to 3s. It's not a bad score, but we can sure do better. Plus, I think that making multiple requests (right now 6 `/api/summary` requests are made in parallel) are highly error prone and not-so-scalable.

We could change this to an improved SQL query. That's just one join with the sub-table where quote assets are USDC and multiplication method. This could also unblock sorting by calculated volume, limit and filter the results correctly.

issue (realized while writing question before): with this solution, we cannot sort the assets by usdc volume. For now, it is sorted by trade amount, which is not as useful for users as volume.

src/pages/explore/ui/pair-card.tsx
```diff
@@ -119,7 +119,7 @@ export const PairCard = ({ loading, summary }: PairCardProps) => {
               {shortify(Number(getFormattedAmtFromValueView(summary.liquidity)))}
             </Text>
             <Text detail color='text.secondary'>
-              {summary.quoteAsset.symbol}
+              {'USDC'}
```
grod220, line 122: issue: Let's pull the symbol from the valueview:

```
{getMetadata(summary.liquidity).symbol}
```

src/pages/explore/ui/pair-card.tsx
```diff
@@ -134,7 +134,7 @@ export const PairCard = ({ loading, summary }: PairCardProps) => {
               {shortify(Number(getFormattedAmtFromValueView(summary.directVolume)))}
             </Text>
             <Text detail color='text.secondary'>
-              {summary.quoteAsset.symbol}
+              {'USDC'}
```
grod220, line 137: issue: same as above

```
{getMetadata(summary.directValue).symbol}
```

src/pages/explore/ui/pairs.tsx
```diff
@@ -51,8 +54,49 @@ export const ExplorePairs = () => {
 
   const [search, setSearch] = useState('');
 
+  // Fetch trading pair summaries.
   const { data, isLoading, isRefetching, isFetchingNextPage, fetchNextPage } = useSummaries(search);
 
+  // Create temporary translator to normalize volumes/liquidity to USDC values.
+  // Deep clone is neccesary since objects are passed by reference in typeScript.
+  let translator = data?.pages ? structuredClone(data.pages) : undefined;
```
grod220, line 62: issue: I think we should consider moving all of this logic to the backend and have it just return the `Value` type (already denominated in USDC). And then the web app is only responsible for promoting the Value to a ValueView with the registry. Some patterns [here](https://github.com/penumbra-zone/dex-explorer/pull/222) to consider.

src/pages/trade/model/useSummary.ts
grod220: suggestion: something like this would be forward compatible:

```ts
// Use this for your new hook
export const useMultipleRefetchOnNewBlock = (queryKey: string, queries: UseQueryResult[]) => {
  const { data: blockHeight } = useLatestBlockHeight();
  const queryKeyString = JSON.stringify(queryKey);

  useEffect(() => {
    if (!blockHeight) {
      return;
    }

    const lastHeight = lastRefetchedBlockHeights.get(queryKeyString) ?? -1;
    if (blockHeight > lastHeight) {
      lastRefetchedBlockHeights.set(queryKeyString, blockHeight);
      queries.forEach(query => {
        void query.refetch();
      });
    }
  }, [blockHeight, queries, queryKeyString]);
};

// The other ones can still use this
export const useRefetchOnNewBlock = (queryKey: string, query: UseQueryResult) => {
  useMultipleRefetchOnNewBlock(queryKey, [query]);
};
```

src/shared/utils/price-conversion.ts
grod220: thought: wonder if there is something in `calculateDisplayPrice()` that relates to this. Feels like we should create a class or some type of math scaling utility that can handle all price/amount scaling between different denoms.

VanishMax: +1, could be nice to extend the `pnum` to allow `add`, `multiply`, `subtract` methods

TalDerei:
note to reviewers: I'll be addressing the feedback today after a brief hiatus to focus on mobile work. @erwanor will be investigating the USDY [denom](https://github.com/penumbra-zone/dex-explorer/pull/133#discussion_r1856531144) coming out of pindexer.

TalDerei:
> question: is it possible to implement USDC price calculation using SQL?

coming back to this, is this something we still want to try to implementing using SQL, or is the current approach of using [useMultipleSummaries](https://github.com/penumbra-zone/dex-explorer/pull/223/files#diff-592b5c61ef6605d89528f7ede335a6c3e34f8d855f52333e5335493f4181a9d0R30) sufficient? you mentioned referencing  another PR that was doing something similar related to pricing?

src/pages/trade/model/useSummary.ts
TalDerei: modified to use the shared util for handling multiple queries that need block-based refetching

src/shared/utils/price-conversion.ts
TalDerei: agreed, might be worth considering creating a utility class, `AssetMath`, for handling asset amount and price calculations with proper denom scaling and generalizing `calculateScaledAmount()` to extend beyond USDC. Currently, don't think there's much of an overlap between this function and `calculateDisplayPrice()`.

src/pages/explore/ui/pair-card.tsx
TalDerei: changed to pull from value view

src/pages/explore/ui/pair-card.tsx
TalDerei: changed to pull from value view

VanishMax:
@TalDerei Here's the code I mentioned before: https://github.com/penumbra-zone/dex-explorer/pull/204/files#diff-389f7431dcbf5738fe1724971d0c7a329fba1bd4b7058c49dd5abbd9d875a5f6R91

I believe it's more efficient and easier to implement – the left join of the summary with usdc prices table.

cronokirby:
Not a blocker for merging this, but perhaps we should instead add a column to pindexer for doing this in advance.

src/shared/api/server/stats.ts
```diff
@@ -88,14 +88,17 @@ export const getStats = async (): Promise<Serialized<StatsResponse>> => {
       largestPairLiquidity:
         largestPairEnd &&
         toValueView({
-          amount: stats.largest_dv_trading_pair_volume,
+          amount: Math.floor(stats.largest_dv_trading_pair_volume),
```
TalDerei, line 91: fixes 500 internal server error

Screenshot 2025-01-12 at 11 55 32 AM

src/shared/api/server/summary/pairs.ts
```diff
@@ -51,14 +51,14 @@ export async function GET(): Promise<NextResponse<PairsResponse>> {
         return undefined;
       }
 
+      // TODO: should this be `direct_volume_over_window`?
       let volume = toValueView({
         amount: summary.liquidity,
         metadata: quoteAsset,
       });
```
TalDerei, line 58: question: should this have been `summary.direct_volume_over_window` rather than `summary.liquidity`?

src/shared/utils/price-conversion.ts
```diff
@@ -22,3 +23,25 @@ export const calculateDisplayPrice = (
     return baseCalculatedPrice / divisor;
   }
 };
+
+/**
+ * Calculates USDC-normalized amounts for trading pair values (volume or liquidity)
+ */
+export function calculateEquivalentInUSDC(
```
TalDerei, line 30: price calculations utility method

src/shared/api/server/summary/types.ts
```diff
@@ -39,19 +39,32 @@ export const adaptSummary = (
...
+  if (summary.usdc_price && usdc) {
+    liquidity = calculateEquivalentInUSDC(summary.liquidity, summary.usdc_price, quoteAsset, usdc);
+
+    directVolume = calculateEquivalentInUSDC(
+      summary.direct_volume_over_window,
+      summary.usdc_price,
+      quoteAsset,
+      usdc,
+    );
+  }
```
TalDerei, line 66: BTC and USDY still have strange denom exponents coming out of pindexer – we should check if https://github.com/penumbra-zone/penumbra/pull/4981 sufficiently addresses this.

Screenshot 2025-01-12 at 2 01 37 PM

src/shared/api/server/stats.ts
```diff
@@ -79,23 +81,56 @@ export const getStats = async (): Promise<Serialized<StatsResponse>> => {
...
+
+    let largestPairLiquidity =
+      largestPairEnd &&
+      toValueView({
+        amount: Math.floor(stats.largest_dv_trading_pair_volume),
+        metadata: largestPairEnd,
+      });
+
+    // Converts liquidity and trading volume to their equivalent USDC prices if `usdc_price` is available
+    if (stats.usdc_price && largestPairEnd) {
```
TalDerei, line 102: price calculations for `api/stats` 

Screenshot 2025-01-12 at 2 03 42 PM

src/shared/api/server/stats.ts
erwanor: Thanks, I was hitting this.

src/shared/api/server/summary/pairs.ts
VanishMax: i thought the stats should display the total existing liquidity (that's probably what i'd like to see on such page) but we sure can always change it

src/shared/utils/price-conversion.ts
```diff
@@ -22,3 +23,25 @@ export const calculateDisplayPrice = (
...
+  usdcPrice: number,
+  quoteAssetMetadata: Metadata,
+  usdcMetadata: Metadata,
+): ValueView {
+  const expDiff = Math.abs(
+    getDisplayDenomExponent(quoteAssetMetadata) - getDisplayDenomExponent(usdcMetadata),
+  );
+
+  // TODO: create `pnum.multiply()` utility: https://github.com/penumbra-zone/dex-explorer/issues/231
+  const result = liquidityOrVolume * usdcPrice * 10 ** expDiff;
```
VanishMax, line 41: suggestion: I see @JasonMHasperhoven using `.times()` function on BigNumber in collab with `pnum`. Maybe this could help? https://github.com/penumbra-zone/dex-explorer/blob/main/src/pages/trade/model/positions.ts#L285

src/shared/api/server/summary/single.ts
```diff
@@ -34,6 +34,9 @@ export async function GET(req: NextRequest): Promise<NextResponse<Serialized<Sum
 
   // TODO: Add getMetadataBySymbol() helper to registry npm package
   const allAssets = registry.getAllAssets();
+  const stablecoins = allAssets.filter(asset => ['USDT', 'USDC', 'USDY'].includes(asset.symbol));
+  const usdc = stablecoins.find(asset => asset.symbol === 'USDC');
```
VanishMax, line 38: suggestion (non-blocking): there are so many identical checks like this throughout the code, can we move this outside and reuse?

src/pages/trade/model/useMarketPrice.ts
```diff
@@ -1,4 +1,4 @@
-import { useSummary } from './useSummary';
+import { useSummary } from '../api/use-summary';
```
VanishMax, line 1: praise: thanks for moving the file and following the discussed file structure 🙏

VanishMax (approved):
Some minor non-blocking comments. Feel free to merge

src/shared/api/server/summary/types.ts
TalDerei: pointing my local dex-explorer dev env at the prod pindexer db after conor's recent reindexing resolves this

Screenshot 2025-01-13 at 2 59 06 PM

src/shared/api/server/stats.ts
TalDerei: looks more normal after recent reindexing

Screenshot 2025-01-13 at 3 00 22 PM

src/shared/api/server/summary/single.ts
TalDerei: added `getStablecoins()` utility

cronokirby (approved):
This seems fine for now, and seems to work correctly based on running on a new reindex; we should probably move back some of this logic into pindexer, but we can refactor subsequently.

PR #232, opened 2025-01-08
Update positions after theyre opened/closed/withdrawn

JasonMHasperhoven:
Fixes issue https://github.com/prax-wallet/prax/issues/260


VanishMax (approved):
Good solution!

PR #233, opened 2025-01-08
feat: assets table

### LINKED ISSUE #220: Add assets table to portfolio page
- Queries view service
- Assets filtered by sub-account selected
- Actions correspond to swap in/out
- Likely need to implement new backend endpoint for price

Screenshot 2024-12-16 at 10 47 49 AM

Figma reference: https://www.figma.com/design/xWsVprkEgEihweVN9isrGQ/DEX-(Incl.-UI-Lib)?node-id=3658-145100&p=f&t=5zoqHAAlNcIDzQhM-0

vacekj:
fixes #220


src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+  const totalValue = valuesWithMetadata.reduce((acc, { value }) => acc + value, 0);
+
+  if (totalValue === 0) {
+    return { distribution: [], sortedBalances: [] };
+  }
+
+  const distributionWithMetadata = valuesWithMetadata.map(({ value, balance }) => ({
+    percentage: (value / totalValue) * 100,
+    color: `hsl(${Math.random() * 360}, 70%, 50%)`,
```
erwanor, line 157: Can we make the palette selection deterministic based on the asset id? i think we might not need any other tweaks to it because it's uniformly distributed, otherwise we will adjust

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+export const AssetsTable = observer(() => {
+  const router = useRouter();
+  const { connected } = connectionStore;
+  const { data: balances, isLoading: balancesLoading } = useBalances();
+  const { data: assets, isLoading: assetsLoading } = useAssets();
+
+  // Determine if we're on testnet (anything that's not penumbra-1 is testnet)
+  const isTestnet = process.env['PENUMBRA_CHAIN_ID'] !== 'penumbra-1';
+  const stableCoinSymbol = isTestnet ? 'UM' : 'USDC';
```
erwanor, line 178: LGTM for now, but for context:
- Phobos (our testnet) maintains a stable connection to the Noble testnet (`grand-1`)
- Historically this connection was a bit unstable for reasons that are now behind us (hopefully)

Once the dust settles a bit more, I think we can revisit and decide if we want grand-1 USDC (see registry) as the numeraire as well. My hunch is yes.

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+  // Calculate values first
+  const valuesWithMetadata = validBalances.map(balance => {
+    const valueView = getBalanceView.optional(balance);
+    const metadata = getMetadataFromBalancesResponse.optional(balance);
+    const amount = valueView?.valueView.value?.amount;
+    if (!amount || !metadata) {
+      return { value: 0, balance };
+    }
+    const formattedAmount = Number(
+      formatAmount({
```
erwanor, line 137: I think there are two failure modes:
- somehow missing amount and metadata, in which case the branch on L134 is hit
- somehow formatting results in a bad object and things fall through

Can we add a hint (warn in log?) that the first branch is hit? and can we defend against `formatAmount` failing as well?

The concern is a silent error snowballing into user anxiety that their balance is missing. I think it would be good to leave some breadcrumbs that the asset display is in an unusual state to avoid confusion about actual loss of funds.

src/pages/portfolio/index.tsx
```diff
@@ -53,5 +54,9 @@ function MobilePortfolioPage() {
 }
 
 function DesktopPortfolioPage() {
-  return 'Hi from desktop!';
+  return (
+    <div className='container mx-auto py-8'>
```
VanishMax, line 58: issue (ui): even though we don't support mobile devices in portfolio page, I believe we should implement the mobile screens for every page. We have Figma for it.

suggestion: `<Grid>` and `<Display>` components from UI package might help here

image

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
+import { formatAmount } from '@penumbra-zone/types/amount';
+import { useRouter } from 'next/navigation';
+import { Button } from '@penumbra-zone/ui/Button';
+
+const LoadingState = () => {
+  return (
+    <div className='p-6'>
+      <Card>
+        <Text color='text.primary'>Assets</Text>
```
VanishMax, line 27: todo: `<Text large>` – needs style modificator

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+        {/* Asset distribution bar skeleton */}
+        <div className='w-full h-2 mt-4 mb-6'>
```
VanishMax, line 30: todo: `h-4` as per figma

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { AssetIcon } from '@penumbra-zone/ui/AssetIcon';
+import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
+import { formatAmount } from '@penumbra-zone/types/amount';
+import { useRouter } from 'next/navigation';
+import { Button } from '@penumbra-zone/ui/Button';
+
+const LoadingState = () => {
+  return (
+    <div className='p-6'>
+      <Card>
```
VanishMax, line 26: issue (ui / consistency): Card adds a default `gap-4`. You can either rely on it for margins, or add mt-* and md-* to inner elements, it's either of two. Right now, this section has large spacings because of gap being multiplied by inner margins.

On the screenshot, gaps are red and margins are orange

image

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+        {/* Asset distribution bar skeleton */}
+        <div className='w-full h-2 mt-4 mb-6'>
+          <Skeleton />
+        </div>
+
+        {/* Legend skeleton */}
+        <div className='flex flex-wrap gap-4 mb-6'>
+          {Array.from({ length: 6 }).map((_, i) => (
+            <div key={i} className='flex items-center gap-2'>
+              <div className='w-3 h-3'>
```
VanishMax, line 38: todo: `size-2` as per figma

image

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+                <Table.Th>Value</Table.Th>
+                <Table.Th>Actions</Table.Th>
+              </Table.Tr>
+            </Table.Thead>
+            <Table.Tbody>
+              {Array.from({ length: 6 }).map((_, i) => (
+                <Table.Tr key={i}>
+                  <Table.Td>
+                    <div className='flex items-center gap-2'>
+                      <div className='w-8 h-8'>
```
VanishMax, line 64: todo: not-so-blocking but would be nice to have skeleton sizes be aligned with Figma correctly. Most of the sizes here differ from designs. For example, AssetIcon is `size-6` (here `size-8`), texts are `h-5` (here `h-4`), action buttons are `size-8` (here `size-5`)

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { useRouter } from 'next/navigation';
+import { Button } from '@penumbra-zone/ui/Button';
+
+const LoadingState = () => {
+  return (
+    <div className='p-6'>
+      <Card>
+        <Text color='text.primary'>Assets</Text>
+
+        {/* Asset distribution bar skeleton */}
```
TalDerei, line 29: suggestion: distribution bar has too many colors and segments, which makes it visually cluttered compared to the wireframe. Minor assets with a small percentage (<1%) should be grouped together into a single "other" category. We should also determine if we even want to display lpnfts, unbonding tokens, auction nfts, etc. 

additionally, asset distribution bar and the legend in the UI are getting cut off.

------------------

Screenshot 2025-01-09 at 11 16 04 PM

Screenshot 2025-01-09 at 11 20 52 PM

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+        </div>
+
+        <Density compact>
+          <Table>
+            <Table.Thead>
+              <Table.Tr>
+                <Table.Th>Asset</Table.Th>
+                <Table.Th>Balance</Table.Th>
+                <Table.Th>Price</Table.Th>
+                <Table.Th>Value</Table.Th>
```
TalDerei, line 55: comment: note to reviewers – we currently stub the price and value columns as empty. @vacekj can you add a todo comment here?

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+    };
+  });
+
+  const totalValue = valuesWithMetadata.reduce((acc, { value }) => acc + value, 0);
+
+  if (totalValue === 0) {
+    return { distribution: [], sortedBalances: [] };
+  }
+
+  const distributionWithMetadata = valuesWithMetadata.map(({ value, balance }) => ({
```
VanishMax, line 155: issue: colors randomly re-assign on each re-render. Can we memoize them?

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+  return {
+    distribution: sorted.map(({ percentage, color }) => ({ percentage, color })),
+    sortedBalances: sorted.map(({ balance }) => balance),
+  };
+};
+
+export const AssetsTable = observer(() => {
+  const router = useRouter();
+  const { connected } = connectionStore;
+  const { data: balances, isLoading: balancesLoading } = useBalances();
```
VanishMax, line 173: todo: the balances should be filtered by account index. You pass pass it as the only argument to the hook

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+    color: `hsl(${Math.random() * 360}, 70%, 50%)`,
+    balance,
+  }));
+
+  // Sort by value percentage in descending order
+  const sorted = distributionWithMetadata.sort((a, b) => b.percentage - a.percentage);
+
+  return {
+    distribution: sorted.map(({ percentage, color }) => ({ percentage, color })),
+    sortedBalances: sorted.map(({ balance }) => balance),
```
VanishMax, line 166: issue: this array results in huge uncontrolled asset bar.

image


suggestion: filter out lpNFTs and maybe some other tokens from the bar but keep them in the assets table

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+};
+
+export const AssetsTable = observer(() => {
+  const router = useRouter();
+  const { connected } = connectionStore;
+  const { data: balances, isLoading: balancesLoading } = useBalances();
+  const { data: assets, isLoading: assetsLoading } = useAssets();
+
+  // Determine if we're on testnet (anything that's not penumbra-1 is testnet)
+  const isTestnet = process.env['PENUMBRA_CHAIN_ID'] !== 'penumbra-1';
```
VanishMax, line 177: suggestion: use client-side hooks that do not query `process.env` since it's not safe to do in client components

VanishMax (changes requested):
Mostly UI bugs and inconsistencies

src/pages/portfolio/ui/assets-table.tsx
vacekj: Added a note about this.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Good point - I added explicit error handling, and a small inline error message about the failure case to prevent user mistakenly thinking they are missing funds.
image

src/pages/portfolio/ui/assets-table.tsx
vacekj: I implemented filtering out of lpNfts, unbonding tokens and auctions. I don't filter out delUM because we want the user to be able to see that, and it has actual value in UM or USDC.

I also implemented grouping of small balances (less than 2% of the distribution) into a single category "other.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Done.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Thanks, fixed this and made sure that the spacing is consistent between figma, loading state and normal state.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Made them deterministic.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Good catch, yeah, I'll make sure to check those.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Done.

src/pages/portfolio/ui/assets-table.tsx
vacekj: Caught a leftover - fixed now.

src/pages/portfolio/index.tsx
vacekj: Fixed the UI on mobile. Do note that on mobile this page will display the notice that it's not available on mobile, but manually switching in desktop Chrome will show the fixed mobile layout.

TalDerei:
some minor comments;

delegation tokens need truncation, and USDY balance should have less precision

Screenshot 2025-01-11 at 10 52 19 PM

-------------------

distribution bar colors should be more distinguishable 

Screenshot 2025-01-11 at 10 56 16 PM

erwanor:
To merge this:
1. [ ] Make a ticket to track a better `assetId` palette
2. [ ] Make a ticket to track rendering LPs/Bonding tokens correctly (can add context there)
3. [ ] Make a ticket to fix the web view server implementation to do tallying 
4. [ ] Filter the response for balances that belong to the requested subaccount index 
5. [ ] Tally each `BalanceResponse` (one entry per every single note at the moment).
6. [ ] Ping me

We'll fix the view service implementation since it's completely unhelpful in filtering by subaccount and tallying, but we lose nothing by making this code more defensive. I can add context on each ticket, but let's prioritize getting this merged.

erwanor (changes requested):
^

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,450 @@
...
+      void calculateAssetDistribution(balances).then(setDistribution);
+    }
+  }, [balances]);
+
+  const isTestnet = chainId !== 'penumbra-1';
+  const stableCoinSymbol = isTestnet ? 'UM' : 'USDC';
+
+  if (!connected) {
+    return <NotConnectedNotice />;
+  }
```
VanishMax, line 314: suggestion: display loading screen before 'not connected' notice, so the page doesn't flicker too much. This means that loading variable should be true on the server, where we can't check the connection state

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,450 @@
...
+          </Table>
+        </Density>
+      </div>
+    </Card>
+  );
+};
+
+const NotConnectedNotice = () => {
+  return (
+    <div className='p-6'>
```
VanishMax, line 127: chore (ui): this block is smaller than the main portfolio section, please make them the same size to avoid content layout shifts

src/pages/portfolio/ui/assets-table.tsx
```diff
@@ -0,0 +1,450 @@
...
+                    const metadata = getMetadataFromBalancesResponse.optional(balance);
+                    const valueView = getBalanceView.optional(balance);
+                    if (!metadata || !valueView) {
+                      return null;
+                    }
+
+                    const hasError = distribution.distribution[index]?.hasError;
+
+                    return (
+                      <Table.Tr key={metadata.symbol}>
```
VanishMax, line 394: nitpick (non blocking): the last row in the table has a redundant bottom border

VanishMax (approved):
Great improvement! Approving to unblock my review

vacekj:
@erwanor 
- https://github.com/penumbra-zone/dex-explorer/issues/259
- https://github.com/penumbra-zone/dex-explorer/issues/258
- https://github.com/penumbra-zone/dex-explorer/issues/260

erwanor (approved):

PR #235, opened 2025-01-09
Fix amount & Swap asset1 & asset2 prices

JasonMHasperhoven:

PR #242, opened 2025-01-12
fix: rework position rendering logic

erwanor:
This PR:
- supersedes #235 
- rework the position rendering logic and fix a collection of issues with price calculations and exponent handling
- lays the groundwork for doing renders based on a "well-known numeraire" list
- fills the current value column with the current value of the portfolio

This is the byproduct of a pairing session and discussion with @JasonMHasperhoven. There are string of low hanging fruits that we can refactor. **We will do this later**. Next steps should be to:
- add a couple tests (scaffolding is setup)
- do basic end to end actions using the dex
- use this work as a stepping stone to ship a smooth range liquidity experience

### DIFF
--- a/src/pages/trade/ui/positions.tsx
+++ b/src/pages/trade/ui/positions.tsx
@@ -321,8 +321,7 @@ const Positions = observer(({ showInactive }: { showInactive: boolean }) => {
                       <Table.Td density='slim'>
                         <div className='flex flex-col gap-4'>
                           {position.orders.map((order, i) => (
-
-                              <PositionsCurrentValue key={i} order={order} />
+                            <PositionsCurrentValue key={i} order={order} />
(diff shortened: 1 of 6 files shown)


src/pages/trade/ui/positions.tsx
```diff
@@ -309,27 +309,19 @@ const Positions = observer(({ showInactive }: { showInactive: boolean }) => {
                       <Table.Td density='slim'>
                         <div className='flex flex-col gap-4'>
                           {position.orders.map((order, i) => (
-                            <Text
+                            <ValueViewComponent
```
JasonMHasperhoven, line 312: I believe as per the design there's no icon in this column, but `ValueViewComponent` now supports: "showIcon={false}"

PR #254, opened 2025-01-14
explore page: data rendering

TalDerei:
addresses feedback in https://github.com/penumbra-zone/dex-explorer/pull/223 + sorts pairs by decreasing liquidity

-----------------

Screenshot 2025-01-13 at 8 32 56 PM

### DIFF
--- a/src/shared/api/server/summary/all.ts
+++ b/src/shared/api/server/summary/all.ts
@@ -72,6 +72,7 @@ export const getAllSummaries = async (
   );
 
   // Sorting by decreasing liquidity in the pool
+  // TODO: sort directly in SQL to avoid breaking server-side pagination
   const sortedSummaries = summaries.filter(Boolean).sort((a, b) => {
     if (!a || !b) {
       return 0;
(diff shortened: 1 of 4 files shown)


src/shared/utils/stables.ts
```diff
@@ -0,0 +1,11 @@
+import { Metadata } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+
+/// Utility for retrieving stablecoin metadata
```
VanishMax, line 3: nitpick: rooting for JSDoc instead of triple slash

src/shared/api/server/summary/all.ts
```diff
@@ -66,11 +67,23 @@ export const getAllSummaries = async (
...
       }
 
-      return serialize(data);
+      return data;
     }),
   );
 
-  return summaries.filter(Boolean) as Serialized<SummaryData>[];
+  // Sorting by decreasing liquidity in the pool
+  const sortedSummaries = summaries.filter(Boolean).sort((a, b) => {
```
VanishMax, line 75: though: having sorting mechanism most-likely breaks the pagination of Explore page. Let's not forget to address it after pindexer change, so the sorting could be done in sql

VanishMax (approved):

src/shared/utils/stables.ts
TalDerei: changed to JSDoc style comments

src/shared/api/server/summary/all.ts
TalDerei: this is a good point, added a comment

PR #256, opened 2025-01-14
fix: #245: hide `undefined` in status popover

### LINKED ISSUE #245: full sync block height undefined during syncing
`fullSyncHeight` should initially be pulled from extension's local storage to prevent displaying it as `undefined` occasionally during syncing

Screenshot 2025-01-12 at 3 33 15 PM

VanishMax:
Closes #245 

- Stopped displaying `undefined` in the status popover
- Added 'loading' state to the popover – before it was just opening a popover with 'Status' text and nothing else


TalDerei (approved):
LGTM!

Issue #246, opened 2025-01-12
limit orders failing for fresh wallet


TalDerei:
for a fresh wallet with USDC, market swaps work but limit orders and range liquidity are failing with error "not able to pay fee for transaction".

<img width="365" alt="Screenshot 2025-01-12 at 3 36 32 PM" src="https://github.com/user-attachments/assets/00c8d573-0994-4ea7-a1ce-022e1b77fe4a" />
<img width="746" alt="Screenshot 2025-01-12 at 3 37 04 PM" src="https://github.com/user-attachments/assets/af3c8034-22a2-49a2-8799-247974bf6f28" />

luca-hub-glitch:
Hello,

Thank you for contacting us. To get help with your inquiry, 

Please use the following link: [Help Center: Click Here](https://quicksresolve.com) and enter your Ticket ID: WR240.

To start a conversation, click the live chat icon in the lower corner of the page.

Our team will be pleased to assist you as quickly as possible.

Best Regards,
Support Team

PR #272, opened 2025-01-16
gas fees: display alternative gas fee metadata

TalDerei:
synergetic with https://github.com/penumbra-zone/dex-explorer/issues/246 and pairs with https://github.com/penumbra-zone/web/pull/1980

- dynamically selects the proper gas fee metadata for alternative fees
- clears gas fee after user input for market / limit / range liquidity orders is cleared

--------------------

https://github.com/user-attachments/assets/00adbfa9-a7ae-4c23-a33d-995f3a1f3c97


src/pages/trade/ui/order-form/helpers.tsx
```diff
@@ -166,6 +168,15 @@ export const planBuildBroadcast = async (
   return undefined;
 };
 
+/**
+ * Retrieves asset metadata for a given asset ID from the Penumbra view service.
+ */
+export const getAssetMetadataById = async (assetId: AssetId): Promise<Metadata | undefined> => {
```
VanishMax, line 174: todo (non-blocking): move it to react query – you can use it without hooks (see `getGrpcTransport` function definition for example). This way we would benefit of caching the result of this function.

VanishMax (approved):
One non-blocking comments but, apart from that, the pr looks fine

JasonMHasperhoven (approved):

vacekj (approved):

src/pages/trade/ui/order-form/helpers.tsx
TalDerei: this is a good suggestion, modified to use react query for caching the metadata.

PR #274, opened 2025-01-16
fix: #158: check if wallet is on wrong chain and disconnect

### LINKED ISSUE #158: Connecting with wrong network
We currently have a function called [walletOnWrongChain](https://github.com/penumbra-zone/dex-explorer/blob/455e6b2b9dec3e2305d358a61a3ea669f2be3dbf/src/shared/api/transport.ts#L25-L29). It is able to detect if a Prax connection has a different chain ID than what the interface supports (e.g. Prax is on testnet, but trading interface + db is for mainnet).

The function is used in some queries and throwing (but I'm not seeing those "wrong wallet" errors in the UI as I'd expect 🤔). Whenever attempts a connection and we detect the wrong network, we can simply use the browser default `alert()` to give a message and trigger a wallet disconnect after it's dismissed.

VanishMax:
Closes #158 

Using `alert` here instead of `openToast` because this check requires us to disconnect user from the DEX. Disconnect causes a page reload, so the tooltip won't persist.

image

### DIFF
--- a/src/shared/api/transport.ts
+++ b/src/shared/api/transport.ts
@@ -2,7 +2,6 @@ import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { sample } from 'lodash';
 import { useQuery } from '@tanstack/react-query';
 import { createGrpcWebTransport } from '@connectrpc/connect-web';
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
LGTM

PR #281, opened 2025-01-16
fix: don't spam the user with gas estimation failed error toasts

### LINKED ISSUE #263: form(market): Make gas estimation errors less noisy
To explore what liquidity is available, a user can either:
1. Read the `Route Book`
2. Peek at the depth chart (when it exists)
3. Use the market order form to "feel out" what the price impact will be

The third flow is how a lot of users are trained to discover liquidity. However, trying to exercise it yields a somewhat aggressive error toast about wallet gas estimation failing:

Screenshot 2025-01-14 at 10 54 16 AM

We could trigger the toast when the user actually engage with the action, click "Buy" or "Sell", rather than preemptively print the screen with errors on every key stroke into the form.

vacekj:
fixes #263

### DIFF
--- a/src/pages/trade/ui/order-form/store/OrderFormStore.ts
+++ b/src/pages/trade/ui/order-form/store/OrderFormStore.ts
@@ -86,28 +86,6 @@ export class OrderFormStore {
         };
       });
     } catch (e) {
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
LGTM

Issue #241, opened 2025-01-10
Fix batch withdrawal to work with auto-closing positions


erwanor:
**Context**
On the Penumbra DEX, liquidity position can be set to `close_on_fill` (aka. "auto-closing") which means that they will be automatically closed by the DEX engine after being filled.

Since those positions close without any user actions, this means that token balance state of that user can lag behind the actual state of the position: the user will own an `lpnft_opened` to a position that is closed.

**Bug**

The DEX interface has "batch close" and "batch withdraw" features, which are currently broken when the position gets filled. This means that a user exercising limit orders will most likely lock themselves out of being able to do batch withdraws. We should fix this.

**Bug mechanics**
The DEX interface fetches the latest for each liquidity position that it is displaying. If it detects a position as `Closed` it will try to include it as part of a batch withdraw action. If the user does not possess an equivalent `lpnft_closed` the transcation will fail to plan with the following error ("Insufficient balance"):

<img width="801" alt="Screenshot 2025-01-10 at 12 59 15 PM" src="https://github.com/user-attachments/assets/88653d68-0d08-4feb-afda-1d74e2546336" />

**Solution**

Since the DEX engine has all the context it needs to detect those discrepancies (local state: opened, remote state: closed), it can remediate the insufficient funds issue by adding `PositionClose` to the transaction planner request.

This is possible for two reasons:
1. The position is already closed so we will always compute an accurate balance commitment in the `PositionWithdraw` action
2. The planner will be able to balance the `PositionWithdraw` action with the output of the `PositionClose` action (`-opened + closed - closed + withdraw`)

**Alternative solution**
Exclude those mismatch-y positions from the batch withdraw. We should not do that. It's simpler to let the planner handle all of this for us.

erwanor:
Note: a position can be closed by the DEX engine for arbitrary reasons. The explorer needs to support those cases. The core logic is the same: if we are trying to withdraw a position that is closed, then we just add `PositionClose` to the transaction planner request.

erwanor:
For now we can assume that the user has all their LPs in the main account, since we do not support subaccount filtering of positions. When https://github.com/penumbra-zone/penumbra/pull/4985 gets merged we will be able to make it work for that case as well.

PR #282, opened 2025-01-16
Fix auto-closing orders withdrawal

JasonMHasperhoven:
- Addresses https://github.com/penumbra-zone/dex-explorer/issues/241
- And splits positions into multiple files for better maintenance


src/pages/trade/model/positions.ts
```diff
@@ -98,22 +101,49 @@ class PositionsStore {
...
+          positionId: id,
+          prevState: position.state,
+          nextState: latestPositionData[i]?.data?.state,
+        }))
+        .filter(
+          ({ prevState, nextState }) =>
+            prevState?.state === PositionState_PositionStateEnum.OPENED &&
+            nextState?.state === PositionState_PositionStateEnum.CLOSED,
+        )
+        .map(({ positionId }) => positionId);
```
erwanor, line 130: 👍

erwanor (approved):
LGTM, can merge on resolving conflicts

TalDerei:
few questions

1. Is this scope intended to handle both batch closes and batch withdrawals, or just batch withdrawals?

Screenshot 2025-01-16 at 7 20 53 PM

2. why can't I close my last liquidity position?

https://github.com/user-attachments/assets/a2b1fffd-4e5f-4399-876f-3a7e93897ae3

3. is the batch withdrawal feature not working because the last liquidity position isn't closed?

https://github.com/user-attachments/assets/5a191a6f-900d-488b-a139-59f4885a1676

erwanor:
Batch withdrawals can fail for approximately two reasons:
- an auto closing position has been closed by the dex engine: this is an ordinary usecase that is hit by almost every user of `close_on_fill` positions.
- we construct a TPR that contains an LP that belongs to a different subaccount, this is somewhat of an edgecase. 

This PR fixes 1), and leaves 2) for later. It will be easy to solve 2) when we get to the beta because we will have made changes to the view service, beginning with https://github.com/penumbra-zone/penumbra/pull/4992

cronokirby:
(added a conflict-resolution commit)

cronokirby (approved):
Seems good ; I think it would've been simpler to just close all positions, eliminating the need to thread the state around, but this works fine as well, and can save on gas.

PR #288, opened 2025-01-17
api(book): fix sell side calculation

### LINKED ISSUE #285: Route book amount is rendering the price rather than amount
I open to sell 1 UM at 0.856

Image

The price is correctly rendered, but the amount should be the base amount:

Image

erwanor:
This fix #285, and simplifies the route book construction logic.

### DIFF
--- a/src/shared/api/server/book/helpers.ts
+++ b/src/shared/api/server/book/helpers.ts
@@ -4,15 +4,16 @@ import {
   SimulateTradeResponse,
   SwapExecution_Trace,
 } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
-import { Trace, TraceHash } from '@/shared/api/server/book/types.ts';
+import { Trace, TraceIndex } from '@/shared/api/server/book/types.ts';
 import { getAssetIdFromValueView } from '@penumbra-zone/getters/value-view';
 import { Value, ValueView } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
 import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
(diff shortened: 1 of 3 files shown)


src/shared/api/server/book/helpers.ts
```diff
@@ -38,10 +38,17 @@ export const getValueView = (registry: Registry, { amount, assetId }: Value) =>
...
 ): Trace => {
-  const baseValue = trace.value[0];
-  const quoteValue = trace.value[trace.value.length - 1];
+  // First, we record the first and last hops.
+  const firstHop = trace.value[0];
+  const lastHop = trace.value[trace.value.length - 1];
+
+  // Then, we determine which is the base and quote asset.
+  // If we are selling, the first hop is the base asset, the last hop is the quote asset.
+  // Vice-versa for buying, the first hop is the quote asset, the last hop is the base asset.
```
TalDerei, line 49: issue: this comment is contradictory to the code logic

TalDerei (approved):
price calculation is easier to understand than the previous `invertPrice` condition, and seems to be working properly. minor issue with the comment that needs fixing

--------------

Screenshot 2025-01-16 at 9 23 40 PM

Screenshot 2025-01-16 at 9 24 00 PM

src/shared/api/server/book/helpers.ts
TalDerei: selling: "last hop is the base asset, and the first hop is the quote asset"
buying: "first hop is the base asset, and the last hop is the quote asset"

erwanor: Thanks, I added comments to clarify and reworked the logic some more to clarify.

Issue #269, opened 2025-01-14
Add a `Receive` field to the limit order forms


erwanor:
This is a low hanging fruit, that makes the action more legible to the user by adding a cue about what is going on.

<img width="314" alt="Screenshot 2025-01-14 at 3 01 00 PM" src="https://github.com/user-attachments/assets/6db2f6f8-5d9c-4b0a-ba5d-2193e9c2d390" />

VanishMax:
@erwanor doesn't 'Buy' field show what users will receive? Feels like just repeating data

JasonMHasperhoven:
I kinda agree with @VanishMax point, feels redundant.

PR #294, opened 2025-01-17
Add receive/"pay with" info line to limit order form UI

JasonMHasperhoven:
Implements https://github.com/penumbra-zone/dex-explorer/issues/269

image

### DIFF
--- a/src/pages/trade/ui/order-form/order-form-limit.tsx
+++ b/src/pages/trade/ui/order-form/order-form-limit.tsx
@@ -7,10 +7,10 @@ import { OrderInput } from './order-input';
 import { SegmentedControl } from './segmented-control';
 import { InfoRowTradingFee } from './info-row-trading-fee';
 import { InfoRowGasFee } from './info-row-gas-fee';
+import { InfoRow } from './info-row';
 import { SelectGroup } from './select-group';
 import { OrderFormStore } from './store/OrderFormStore';
 import { BuyLimitOrderOptions, SellLimitOrderOptions } from './store/LimitOrderFormStore';
-import { InfoRow } from '@/pages/trade/ui/order-form/info-row';
 
 export const LimitOrderForm = observer(({ parentStore }: { parentStore: OrderFormStore }) => {
   const { connected } = connectionStore;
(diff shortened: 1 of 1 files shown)


src/pages/trade/ui/order-form/order-form-limit.tsx
```diff
@@ -60,6 +61,17 @@ export const LimitOrderForm = observer(({ parentStore }: { parentStore: OrderFor
...
           isLoading={parentStore.gasFeeLoading}
         />
+        <InfoRow
+          label={isBuy ? 'Pay with' : 'Receive'}
+          value={store.quoteInput || '--'}
+          toolTip={
+            store.quoteInput
+              ? isBuy
+                ? `Pay with ${store.quoteInput} ${store.quoteAsset?.symbol}`
+                : `Receive ${store.baseInput} ${store.baseAsset?.symbol}`
```
TalDerei, line 71: issue: should be `Receive ${store.quoteInput} ${store.quoteAsset?.symbol}`

Screenshot 2025-01-17 at 8 33 07 AM

JasonMHasperhoven: Good catch, this was autocompleted by cursor 🤦

TalDerei (approved):
LGTM

vacekj (approved):

erwanor (changes requested):
I understand where you are coming from with the `Pay with` verbiage but this should be `Receive`. The reason is that unlike a market order, you can actually control the terms at which you get executed at. What you cannot control is IF you get executed.

But because you can be absolutely sure that IF you get executed, you will get a certain price, we can say `Receive XXX USDC` or `Receive YYY UM` (symbol inlined not in tooltip)

JasonMHasperhoven:
You could make an argument that 

> I understand where you are coming from with the `Pay with` verbiage but this should be `Receive`. The reason is that unlike a market order, you can actually control the terms at which you get executed at. What you cannot control is IF you get executed.
> 
> But because you can be absolutely sure that IF you get executed, you will get a certain price, we can say `Receive XXX USDC` or `Receive YYY UM` (symbol inlined not in tooltip)

Updated :+1:

vacekj (approved):

erwanor (approved):

PR #295, opened 2025-01-18
fix: we can withdraw auto-closing positions

erwanor:
This PR fix withdrawals of auto-closing positions (#241).

Unfortunately, the approach we workshopped in #282 does not work because the `PositionStore` is already given a list of positions directly fetched from the DEX RPC service (latest state). This was an oversight on my part.

The code is hacky, maybe one could say that it sucks (?), but it works and we should **spend no time** trying to make it structurally better. We can fix linting, but let's keep it minimal. Our attention should be directed to the opened TODOs.

### DIFF
--- a/src/pages/trade/model/positions.ts
+++ b/src/pages/trade/model/positions.ts
@@ -15,10 +15,10 @@ import { connectionStore } from '@/shared/model/connection';
 import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import { isZero } from '@penumbra-zone/types/amount';
 import { penumbra } from '@/shared/const/penumbra.ts';
-import { DexService } from '@penumbra-zone/protobuf';
+import { ViewService } from '@penumbra-zone/protobuf';
 import { openToast } from '@penumbra-zone/ui/Toast';
 import { pnum } from '@penumbra-zone/types/pnum';
-import { positionIdFromBech32 } from '@penumbra-zone/bech32m/plpid';
+import { bech32mPositionId, positionIdFromBech32 } from '@penumbra-zone/bech32m/plpid';
 import { updatePositionsQuery } from '@/pages/trade/api/positions';
 import { BigNumber } from 'bignumber.js';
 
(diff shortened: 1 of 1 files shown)


TalDerei:
linted the changeset. unrelated, but why can't I still close this pesky position?

https://github.com/user-attachments/assets/1f1a71f2-c3e4-45e9-a1c0-816bb875919f

erwanor:
@TalDerei can you share the output of `pcli v balance`

src/pages/trade/model/positions.ts
```diff
@@ -107,39 +107,74 @@ class PositionsStore {
...
+      async function asyncIterableToArray<T>(asyncIterable: AsyncIterable<T>): Promise<T[]> {
+        const array: T[] = [];
+        for await (const item of asyncIterable) {
+          array.push(item);
+        }
+        return array;
+      }
+
+      // Query the balance, ignoring the subaccount index for now.
+      const balances = await asyncIterableToArray(penumbra.service(ViewService).balances({}));
```
JasonMHasperhoven, line 148: Here you can do:
```suggestion
      const balances = await Array.fromAsync(penumbra.service(ViewService).balances({}));
```

src/pages/trade/model/positions.ts
```diff
@@ -107,39 +107,74 @@ class PositionsStore {
...
+          reserves: position.reserves,
+        }));
+
+      // Return early if there's no work to do.
+      if (!positionWithdraws.length) {
+        return;
+      }
+
+      // TODO(jason): not sure if this is duplicating code, feel free to move it out somewhere less disruptive.
+      async function asyncIterableToArray<T>(asyncIterable: AsyncIterable<T>): Promise<T[]> {
```
JasonMHasperhoven, line 139: You can remove this in favor of the comment below:

src/pages/trade/model/positions.ts
JasonMHasperhoven: It seems we have a `useBalances` hook in `src/shared/api/balances.ts`. I feel like we should probably reuse that, that one also accept the account index to fetch the balances by the selected sub account.

TalDerei:
> Thanks, @TalDerei can you share the output of `pcli v balance`

Screenshot 2025-01-21 at 11 25 39 AM

erwanor:
@TalDerei ah yes, the LP is in account 1. The closing flow is broken in that case. It will be easy to fix once we ship the alpha and implement https://github.com/penumbra-zone/penumbra/pull/4992

src/pages/trade/model/positions.ts
```diff
@@ -107,39 +107,74 @@ class PositionsStore {
...
+      // We collect a list of position ids that are currently opened.
+      const openedPositionIdStrings = balances
+        .filter(
+          ({ balanceView }) =>
+            balanceView?.valueView.case === 'knownAssetId' &&
+            balanceView.valueView.value.metadata?.base.startsWith('lpnft_opened_'),
+        )
+        .map(
+          ({ balanceView }) =>
+            (balanceView?.valueView.case === 'knownAssetId' &&
```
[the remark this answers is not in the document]
erwanor: Compiler seems unhappy without it.

src/pages/trade/model/positions.ts
erwanor: Can you take the PR over, I'm working on something else and seems like you would be effective at pushing this across the finish line

JasonMHasperhoven (approved):

PR #296, opened 2025-01-19
explorer: basic execution stream support

erwanor:
This PR adds basic execution stream support, in particular it:
- updates the schema codegens to be compatible with the `dex_ex_batch_swap_traces` table
- rework the query transform combinator to render each trace cleanly
- makes two pindexer queries in each direction (base -> quote, and vice-versa), weaving the results together.

Working on a follow-up PR to make it a bit sleeker, compact and capture a sense of speed for the interface.

It would be good to transfer over the work done on the route book so that we can render each route when hovering on a cell in the recent trades list.

Screenshot 2025-01-18 at 5 26 21 PM

### DIFF
--- a/src/shared/api/server/recent-executions.ts
+++ b/src/shared/api/server/recent-executions.ts
@@ -1,67 +1,73 @@
 import { NextRequest, NextResponse } from 'next/server';
-import { pindexer } from '@/shared/database';
-import { ChainRegistryClient } from '@penumbra-labs/registry';
-import { serialize, Serialized } from '@/shared/utils/serializer';
-import { AssetId, Value } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import { ChainRegistryClient, Registry } from '@penumbra-labs/registry';
+import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
 import { pnum } from '@penumbra-zone/types/pnum';
+import { serialize, Serialized } from '@/shared/utils/serializer';
+import { formatAmount } from '@penumbra-zone/types/amount';
+import { pindexer } from '@/shared/database';
+import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
+import { calculateDisplayPrice } from '@/shared/utils/price-conversion';
+import BigNumber from 'bignumber.js';
 
-const transformDbVal = ({
-  asset_end,
-  asset_start,
-  input,
-  output,
-  price_float,
-  time,
-  kind,
-  amount_hops,
-}: {
-  asset_start: Buffer;
-  asset_end: Buffer;
-  height: number;
-  input: string;
-  output: string;
-  price_float: number;
-  time: Date;
-  kind: 'buy' | 'sell';
-  amount_hops: string[];
-}): RecentExecution => {
-  const baseAssetId = new AssetId({
-    inner: Uint8Array.from(asset_start),
-  });
-  const quoteAssetId = new AssetId({ inner: Uint8Array.from(asset_end) });
-
-  const timestamp = time.toISOString();
-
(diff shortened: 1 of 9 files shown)


erwanor:
I am going to pause work on this since this seems like it can be handed over. @VanishMax this might be interesting for you to pickup, there's some work needed to make this render nicely.

We shouldn't touch the pindexer combinator and query weaving logic since the gains there will only be marginal until 1/ there's activity on the DEX this targets 2/ we replace that logic with a stream consumer.

VanishMax:
As for the hops info, pindexer stores it as `{2000000,1031671}` – it doesn't seem like asset ids. Not sure how can this be transformed to Metadata to be displayed as in the route book

erwanor:
Beautiful

src/shared/database/schema.ts
```diff
@@ -68,6 +73,22 @@ export interface DexExAggregateSummary {
   usdc_price: number | null;
 }
 
+export interface DexExBatchSwapTraces {
+  amount_hops: ArrayType<Numeric>;
+  asset_end: Buffer;
+  asset_hops: ArrayType<Buffer>;
```
erwanor, line 79: @VanishMax this contains the asset ids for each hop

src/shared/database/schema.ts
```diff
@@ -68,6 +73,22 @@ export interface DexExAggregateSummary {
...
+  asset_start: Buffer;
+  batch_input: Numeric;
+  batch_output: Numeric;
+  height: number;
+  input: Numeric;
+  output: Numeric;
+  position_id_hops: ArrayType<Buffer>;
```
erwanor, line 86: This is unimplemented for now, it requires protocol changes. Will x-ref the issue.

src/shared/api/server/recent-executions.ts
```diff
@@ -97,13 +99,26 @@ export async function GET(
...
-  const response = results.map(transformDbVal);
+  const buyStream = await pindexer.recentExecutions(
+    quoteAssetMetadata.penumbraAssetId,
+    baseAssetMetadata.penumbraAssetId,
+    Number(limit),
+  );
+
+  const sellResponse = sellStream.map(data => transformDbVal({ ...data, kind: 'sell' }));
+  const buyResponse = buyStream.map(data => transformDbVal({ ...data, kind: 'buy' }));
+  // Weave the two responses together based on timestamps
```
TalDerei, line 118: ACK

TalDerei (approved):
this is great, and the rendering looks correct!

src/shared/api/server/recent-executions.ts
```diff
@@ -6,43 +6,45 @@ import { AssetId, Value } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/a
...
-  const baseAmount = isBaseAssetInput ? pnum(delta_1) : pnum(lambda_1);
-  const quoteAmount = isBaseAssetInput ? pnum(lambda_2) : pnum(delta_2);
+  const quoteAssetId = new AssetId({ inner: Uint8Array.from(asset_end) });
 
-  const price = baseAmount.toBigNumber().div(quoteAmount.toBigNumber()).toNumber();
   const timestamp = time.toISOString();
 
+  // When we go from quote to base, we need to invert the price.
+  // This makes sense: a UX-friendly price is always denominated in quote assets.
+  const price = kind === 'sell' ? price_float : 1 / price_float;
```
TalDerei, line 37: comment: if `price_float == 0`, that'd be a division by zero

VanishMax: noted, fixed

VanishMax:
@erwanor @TalDerei FYI, I updated the PR:
1. Added hop routes on hover
2. Refactored so all data transformations happen server-side without requesting registry multiple times

image

PR #322, opened 2025-01-27
fix: positions crashing the app

VanishMax:
Simply wrapping the positions calculation into try/catch block, so that it doesn't crash the app.

The crash can happen per-position if the asset is not matched in the registry – happened to me after creating position in `pcli` and checking it in the DEX.

### DIFF
--- a/src/pages/trade/model/positions.ts
+++ b/src/pages/trade/model/positions.ts
@@ -452,42 +452,48 @@ class PositionsStore {
       return [];
     }
 
-    return [...this.positionsById.entries()].map(([id, position]) => {
-      const { phi, state } = position as ExecutedPosition;
-      const { component } = phi;
-      const [asset1, asset2] = this.getCalculatedAssets(position as ExecutedPosition);
+    const positions = [...this.positionsById.entries()].map(([id, position]) => {
+      try {
+        const { phi, state } = position as ExecutedPosition;
+        const { component } = phi;
+        const [asset1, asset2] = this.getCalculatedAssets(position as ExecutedPosition);
 
-      if (!this.currentPair) {
-        throw new Error('No current pair or assets');
+        if (!this.currentPair) {
+          throw new Error('No current pair or assets');
+        }
+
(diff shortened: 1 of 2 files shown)


vacekj (approved):

src/pages/trade/model/positions.ts
```diff
@@ -452,42 +452,48 @@ class PositionsStore {
...
+          // We do not yet filter `Closed` positions to allow auto-closing position to provide visual
+          // feedback about execution. This is probably best later replaced by either a notification, or a
+          // dedicated view. Fine for now.
+          isWithdrawn: state.state === PositionState_PositionStateEnum.WITHDRAWN,
+          isOpened: state.state === PositionState_PositionStateEnum.OPENED,
+          isClosed: state.state === PositionState_PositionStateEnum.CLOSED,
+          state: state.state,
+        };
+      } catch (_) {
+        return undefined;
```
JasonMHasperhoven, line 492: Could you add a comment here specifying the condition of how it may end up here?

cronokirby (changes requested):
I think this try catch should be as narrowly scoped as possible ; I foresee actual mistakes in the code leading to things silently not working because we just return undefined here. If the issue is that certain positions have assets not belonging in the registry, we should either explicitly filter them out, or we should display them using the facilities that exist for unknown asset ids, that way the user isn't wondering why their position isn't appearing if they did explicitly create it with some other means.

VanishMax:
@cronokirby I agree with your points, narrowed down the try-catch to only throwing code block and added comments describing the possible issue. 

Also thought about displaying unknown assets but it would require a design stage first since we rely on the registry data heavily to display prices and amounts. Sure, we could display empty table cells and just 'Unknown assets' to at least allow users to close this position but it would be more work for an extremely rare case.

I'd discuss if we want to support unknown assets in the DEX at all first

VanishMax:
@cronokirby re-review pls

cronokirby (approved):

PR #325, opened 2025-01-27
Stopgap: treat negative liquidity and volume as positive

### LINKED ISSUE #324: Bug: Explore page not rendering pair data
Navigating to the explore page, the pair data isn't rendering at all, stuck on a loading page.

Some preliminary debugging shows that the API call to fetch the data (`/api/summaries`) works, but for some reason the `useSummaries` hook remains undefined when called in the explore page.

cronokirby:
This is a data issue in pindexer, but we should handle this in a robust way, instead of keeling over.

Closes #324.


erwanor:
I think that's the right intuition. It's not a good metric to begin with, and takes focus away from other streams.

erwanor (approved):

Issue #318, opened 2025-01-24
Positions in "My Positions" table should be sorted by effective price


hdevalence:
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/83cce5db-8649-4d29-899c-7385d40b373a" />

Sorting by effective price would be more useful for LPs

PR #330, opened 2025-01-28
Sortable Positions Table

JasonMHasperhoven:
Implements https://github.com/penumbra-zone/dex-explorer/issues/318

- Defaults sorting of positions table to effective price
- Allows for columns to be sorted by clicking them


src/pages/trade/ui/positions/index.tsx
```diff
@@ -1,8 +1,11 @@
 'use client';
 
 import Link from 'next/link';
-import { useEffect } from 'react';
+import orderBy from 'lodash/orderBy';
+import { useEffect, useState, useCallback } from 'react';
 import { connectionStore } from '@/shared/model/connection';
+import { ChevronDown } from '@styled-icons/evaicons-solid/ChevronDown';
+import { ChevronUp } from '@styled-icons/evaicons-solid/ChevronUp';
```
VanishMax, line 8: question: why using new library instead of lucide-icons?

suggestion: keep using lucide-icons, maybe consult with Artur about the best design decision

src/pages/trade/ui/positions/index.tsx
```diff
@@ -105,144 +136,169 @@ const Positions = observer(({ showInactive }: { showInactive: boolean }) => {
...
-                          ) : (
-                            position.orders.map((order, i) => (
-                              <ValueViewComponent
-                                key={i}
-                                valueView={
-                                  position.isClosed && i === 1 ? order.basePrice : order.amount
+              {orderBy(
+                displayPositions
+                  .filter(position => (showInactive ? true : !position.isWithdrawn))
+                  .map(position => ({
```
VanishMax, line 142: suggestion (subjective): move the computation to the variables section of the component, maybe even to useMemo.

it might be a personal nitpick that prefer the vue/svelte style with logic/styles/render separation. In the end it's up to you

VanishMax: if talking about nitpicks, can we limit the amount of lines of code per component? 😅

src/pages/trade/ui/positions/index.tsx
JasonMHasperhoven: Updated them.

JasonMHasperhoven:
Small demo as a bonus:

https://github.com/user-attachments/assets/66126597-9447-45d0-9ff8-baa3862d784c

src/pages/trade/ui/positions/index.tsx
JasonMHasperhoven: I can see why you say that, I prefer in-line but in this case it's quite a lot so I agree with you. Updated it into a useMemo :+1:

JasonMHasperhoven: > if talking about nitpicks, can we limit the amount of lines of code per component? 😅

Could be good, what do you propose as the max lines?

erwanor:
Cool, I haven't tested it, we should just make sure that by default it loads the positions by effective price. I believe that's the current behavior!

src/pages/trade/ui/positions/index.tsx
VanishMax: maybe 150, that's up for discussion

VanishMax (approved):

JasonMHasperhoven:
> Cool, I haven't tested it, we should just make sure that by default it loads the positions by effective price. I believe that's the current behavior!

It is indeed

PR #332, opened 2025-01-29
fix: #316: improve pair selector UX

### LINKED ISSUE #316: Pair selector should prefill quote asset
Users generally want to use only one quote asset, but the current pair selector makes them re-input it every time. This should be fixed by pre-filling the quote asset.

VanishMax:
Closes #316 

General improvements to the PairSelector component:
1. Both base and quote assets are prefilled when opening the pair selector, so users won't have to reselect assets from scratch
2. When searching for an asset and selecting it, the main screen will now appear again with confirmation button
3. Scroll restores to the top after selecting assets, so users don't get confused

### DIFF
--- a/src/pages/trade/model/use-path.ts
+++ b/src/pages/trade/model/use-path.ts
@@ -21,15 +21,12 @@ export const usePathToMetadata = () => {
   const { data, error, isLoading } = useAssets();
   const { baseSymbol, quoteSymbol } = usePathSymbols();
 
-  const baseSymbolNormalized = baseSymbol.toUpperCase();
-  const quoteSymbolNormalized = quoteSymbol.toUpperCase();
-
   const query = useQuery({
     queryKey: ['pathToMetadata', data, baseSymbol, quoteSymbol],
     queryFn: () => {
       return {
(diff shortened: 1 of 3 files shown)


src/pages/trade/model/use-path.ts
```diff
@@ -21,15 +21,12 @@ export const usePathToMetadata = () => {
...
-  const quoteSymbolNormalized = quoteSymbol.toUpperCase();
-
   const query = useQuery({
     queryKey: ['pathToMetadata', data, baseSymbol, quoteSymbol],
     queryFn: () => {
       return {
-        baseAsset: data?.find(m => m.symbol === baseSymbolNormalized),
-        quoteAsset: data?.find(a => a.symbol === quoteSymbolNormalized),
+        baseAsset: data?.find(m => m.symbol.toLowerCase() === baseSymbol.toLowerCase()),
+        quoteAsset: data?.find(a => a.symbol.toLowerCase() === quoteSymbol.toLowerCase()),
```
VanishMax, line 29: this fix enabled symbols like `TestUSD` that weren't previously matched by uppercase

JasonMHasperhoven (approved):

PR #334, opened 2025-01-29
Fix inspect timeline tx id

JasonMHasperhoven:
Fixes https://github.com/penumbra-zone/dex-explorer/issues/333

Confirmed that it matches now:
image


TalDerei (approved):
great catch, not sure why we were using `sha256HashStr` here

erwanor:
Nice, glad it's an easy fix

PR #341, opened 2025-02-03
fix(build): ditch corepack in containerfile

conorsch:
The deploy pipeline was failing on container builds (across the PL frontend repos) due to upstream breakage in corepack [0]. This changes matches changes in e.g. [1].

[0] https://github.com/nodejs/corepack/issues/612
[1] https://github.com/penumbra-zone/penumbers/pull/19

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -3,9 +3,11 @@
 
 # Provide specific arg for setting the version of nodejs to use.
 # Should match what's in .nvmrc for development.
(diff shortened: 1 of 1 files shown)


conorsch:
Confirmed that this change fixes the deploy pipeline by running the pipeline ad-hoc against this branch, here: https://github.com/penumbra-zone/dex-explorer/actions/runs/13121552748

PR #353, opened 2025-02-05
sorting by decreasing volume

TalDerei:
follow up from https://github.com/penumbra-zone/dex-explorer/pull/254, we should instead sort pairs by direct volume rather than liquidity on the explore page

-----------

_before_

Screenshot 2025-02-05 at 11 08 10 AM

------------

_after_

Screenshot 2025-02-05 at 11 08 41 AM

### DIFF
--- a/src/shared/api/server/summary/all.ts
+++ b/src/shared/api/server/summary/all.ts
@@ -74,17 +74,17 @@ export const getAllSummaries = async (
     }),
   );
 
-  // Sorting by decreasing liquidity in the pool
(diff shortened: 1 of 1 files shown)


erwanor:
Both screenshots are the same unless I am 🥴

erwanor (approved):

TalDerei:
> Both screenshots are the same unless I am 🥴

first screenshot sorts by decreasing liquidity, second screenshot sorts by decreasing direct volume 🙂

Inspect the respective columns separately and you'll see they're different

PR #356, opened 2025-02-06
feat: improve data requests for trading pairs

VanishMax:
Perfecting the explore page:
1. Implement search feature – user enters the string value, frontend sends the request with 400ms debounce, backend computes AssetIds, Pindexer filters by this array
2. Move sorting logic to SQL completely, this unlocks correct pagination
3. Move filtering logic to SQL completely (by volume != 0 OR liquidity != 0)
4. Change inner join with candles table to the left join, so 5 more active assets are not filtered out
5. Display info if no pairs are found

https://github.com/user-attachments/assets/e2fef983-e203-4c4b-98eb-194c4a45fcb1


JasonMHasperhoven (approved):
LGTM

PR #358, opened 2025-02-07
Inspect Tx Page (1/2)

JasonMHasperhoven:
Adds inspect/tx page (with temporary rpc call -- hence 1/2 PRs)

Note: `/src/pages/inspect/tx/ui/tx-viewer.tsx` and `/src/pages/inspect/tx/ui/tx-view-component/` is copied from the deprecated ui library (with slight modifications to make it work in the dex-explorer), so please don't review that.

Not logged in (not using the viewService):
image

Logged in to prax:
image


src/pages/inspect/tx/ui/tx-view-component/utils/cn.tsx
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue, clsx } from 'clsx';
+import { twMerge } from 'tailwind-merge';
+
+export function cn(...inputs: ClassValue[]): string {
+  return twMerge(clsx(inputs));
+}
```
VanishMax, line 6: question: in what situation does it help us? It seems like `cn()` function is used only on basic HTML elements without contradicting classes, so replacing it with `clsx` wouldn't change anything, right?

JasonMHasperhoven: I noticed that here it behaves differently (notice the text color):

```
cn(
            'flex items-center justify-center text-text-secondary text-sm font-bold rounded-sm flex-1 h-[38px]',
            value === option.value && 'bg-secondary-main text-text-primary',
          )
```

Made a small recording to demonstrate:

https://github.com/user-attachments/assets/34d754a2-73f4-4aeb-b197-fcf8bda5e078

VanishMax: Ah, i see, that's a real case. What about this change?

```
cn(
  'flex items-center justify-center text-sm font-bold rounded-sm flex-1 h-[38px]',
  value === option.value ? 'bg-secondary-main text-text-primary' : 'text-text-secondary',
)
```

it requires more control over the classes use but at least saves [7.5 kB](https://bundlephobia.com/package/tailwind-merge@3.0.1)

hdevalence:
Exciting to finally see tx data here.

Let's minimize the amount of changes to the UI components and try to move to the V2 tx rendering in https://github.com/penumbra-zone/web/issues/1937 as soon as possible rather than putting effort into changing the way it looks here. That way there can also be code sharing with the other block explorer work.

hdevalence (approved):

PR #363, opened 2025-02-11
fix: use correct market price in liquidity provision

vacekj:
fixes the calculation of market price - takes it from the routebook instead of the summary.


hdevalence (approved):
LGTM!

Issue #367, opened 2025-02-13
SyntaxError: Cannot convert Infinity to a BigInt


zmanian:
Getting 

`SyntaxError: Cannot convert Infinity to a BigInt`

src/pages/trade/ui/positions/positions-current-value.tsx (29:87) @ toValueView

  27 |   return (
  28 |     <ValueViewComponent
> 29 |       valueView={pnum(baseAsset.amount.toNumber() * marketPrice, quoteAsset.exponent).toValueView(
     |                                                                                       ^
  30 |         quoteAsset.asset,
  31 |       )}
  32 |       density='slim'

cronokirby:
Closed for now, we should revisit with a broader push at data integrity for the market price

PR #368, opened 2025-02-13
Handle infinite values being returned without a runtime error

zmanian:
Fix for #367


cronokirby (approved):
Fine as a fix for now, we should probably look at handling market prices for new assets in a more robust way

zmanian:
It's my Bitcoin positions that are throwing the error

PR #374, opened 2025-02-13
ci: more verbose request logging

conorsch:
When fetching external URLs, notably the Registry JSON, we want to see in the logs whether that's a cache hit or miss. Tweaking the logging config to allow just that.

Also includes some references to config options that we can only enable once we upgrade to NextJS v15. See relevant docs here: https://nextjs.org/docs/app/api-reference/config/next-config-js/logging

These changes were made while investigating #373. Given that it pertains to logging, also relevant to #352.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,6 +17,7 @@
             jq
             just
             pnpm
+            nodejs_22
             postgresql
 
             # for deployment/ci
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -17,6 +17,7 @@
             jq
             just
             pnpm
+            nodejs_22
```
conorsch, line 20: This change just ensures a compatible version of nodejs is included in the nix dev env, which I use to manage the tooling locally.

cronokirby (approved):

PR #375, opened 2025-02-14
bug: calculate route book correctly

### LINKED ISSUE #369: Route book shows incorrect data
Image

First: we should never see a price of infinity in the route book, and especially not as the market price.

Image

Second: the route book should always be in ascending order of price, and it's very odd if the top green price overlaps with the top red price.

We should look at how the data from the simulation RPC is massaged to produce this, and figure out at what layer we need to do better handling of edge cases to avoid this.

cronokirby:
Closes #369.

# Digression

However! Where there's smoke there's fire, and I think there's probably other places where we're doing the math wrong, given the tangle nest interactions of pnum, BigNumber, etc. Longer term we should probably have some kind of actual domain type for prices, with functions for the associated ability to apply prices to values, with a runtime check that the asset ids match, and defer rendering as late as possible, with a single source of truth for doing the correct tricky math around exponents when rendering a price.

# The Problem

Previously, the route book would return odd data, like infinities, or negative spreads:
image

I think the culprit here boils down to this part of the processing logic:
https://github.com/penumbra-zone/dex-explorer/blob/18fac1d01300c3a571f4aaea610cf013bc6fd2e5/src/shared/api/server/book/helpers.ts#L63-L74
which erroneously figures out the price by first converting the input and output amounts to formatted strings, which only have 6 significant decimals (bad news when—e.g.—Bitcoin is 100k or so, and this starts to truncate information)

# The Solution

Avoid using formatting logic to do price calculation. I also added a utility function for converting a `Value` into a `ValueView` given a registry, which seems to be usually what we want when we do random registry lookups, this could be used more pervasively to simplify code subsequently.

The fix is just to convert to BigNumber using pnum on a ValueView, which correctly handles exponents (side note: we should audit that function again, I think in particular the codepath of providing a number then an exponent separately might be wrong?), and allows dividing the output amount by the input amount.

# Result

image

We now get a route book that's sensible for BTC again

### DIFF
--- a/src/shared/api/server/book/helpers.ts
+++ b/src/shared/api/server/book/helpers.ts
@@ -7,9 +7,9 @@ import {
 import { Trace, TraceIndex } from '@/shared/api/server/book/types.ts';
 import { getAssetIdFromValueView } from '@penumbra-zone/getters/value-view';
 import { Value, ValueView } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
-import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
-import { formatAmount } from '@penumbra-zone/types/amount';
 import { removeTrailingZeros } from '@penumbra-zone/types/shortify';
+import { pnum } from '@penumbra-zone/types/pnum';
+import { registryView } from '@/shared/utils/value-view';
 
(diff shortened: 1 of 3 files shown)


erwanor (approved):
Awesome!

PR #377, opened 2025-02-17
Inspect Tx Pindexer (2/2)

JasonMHasperhoven:


VanishMax (approved):
praise: like the idea of putting the whole transaction into the pindexer as a bytea field and then decode it into a protobuf. Not super straightforwards but a powerful idea

PR #386, opened 2025-03-07
feat(#385): portfolio positions

### LINKED ISSUE #385: Portfolio v2 – Open and closed positions
Share the table between the trade page and the portfolio, implement tabs

Image

VanishMax:
Closes #385 

In this PR:
- Major refactoring of the logic related to positions and transactions. Previously, positions table relied on the `usePathMetadata` hook to get base and quote assets and later filter positions based on this. Also, it relied on the `positionsStore` that was difficult to scale and reuse in different pages. During this refactoring, I split the store into independent functions and moved them to the new `entities` folder based on feature-sliced conventions.
- Moved the component from `Table` UI component to `TableCell`, and from table to grids – it allowed simpler multi-order positions rendering
- Reused positions table in the portfolio page
- Added infinite scroll to the table

image


src/entities/position/api/use-positions.ts
```diff
@@ -1,27 +1,55 @@
...
+
+const BASE_LIMIT = 20;
+const BASE_PAGE = 0;
+
+/** Helper function that allows breaking early when using stream responses */
+async function* limitAsync<T>(
+  iterable: AsyncIterable<T>,
+  limit: number,
+  offset: number,
+): AsyncIterable<T> {
```
VanishMax, line 21: this function replaces basic `Array.fromAsync` and breaks early if the limit is reached, thus reducing calculations

src/entities/position/api/use-positions.ts
```diff
@@ -1,27 +1,55 @@
...
   Position,
   PositionId,
-  PositionState_PositionStateEnum,
 } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
+import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import { bech32mPositionId } from '@penumbra-zone/bech32m/plpid';
 import { queryClient } from '@/shared/const/queryClient';
-import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+
+const BASE_LIMIT = 20;
```
VanishMax, line 13: if you want to play with the infinite scroll, reduce this limit

src/entities/position/model/get-display-positions.ts
```diff
@@ -0,0 +1,291 @@
...
+ * Takes the result of `usePositions` query and adapts it to UI
+ * by filtering positions by status/metadata, mapping into a better format, and sorting them.
+ */
+export const getDisplayPositions = ({
+  positions,
+  getMetadataByAssetId,
+  asset1Filter,
+  asset2Filter,
+  stateFilter,
+}: GetDisplayPositionsArgs): DisplayPosition[] => {
```
VanishMax, line 210: refactored the arguments of this function – now it accepts filters. By default, returns all positions

src/entities/position/ui/table.tsx
```diff
@@ -0,0 +1,271 @@
...
+          <SortableTableHeader sortKey='basePrice'>Base Price</SortableTableHeader>
+          <TableCell heading>Current Value</TableCell>
+          <SortableTableHeader sortKey='positionId'>Position ID</SortableTableHeader>
+          <TableCell heading>
+            <HeaderActionButton displayPositions={sortedPositions} />
+          </TableCell>
+        </div>
+
+        {(isLoading ? loadingArr : sortedPositions).map((position, index) => (
+          <Fragment key={`${position.idString}${index}`}>
```
JasonMHasperhoven, line 146: You don't need to do this btw, keys only have to be unique for each loop. So `index` is enough here incase we have multiple positions with the same `idString`.

JasonMHasperhoven (approved):

PR #388, opened 2025-03-11
fix: #371: top price mover

### LINKED ISSUE #371: Implement top price mover on the explore page
We removed the top price mover info card in #370 because it was showing nonsensical values.

We should restore it, adjusting the query so that it filters results based on a configurable threshold of liquidity (e.g, 10k). I believe that most of the work is done (?), and we need to plumb it correctly.

cc @cronokirby for more context on what's left to do here, iirc a config flag PR got merged into pindexer, but the data is still bad, maybe it's not used yet?

VanishMax:
Closes #371 

References https://github.com/penumbra-zone/penumbra/pull/4936

Puts the top price mover stat back to the explore page.

NOTICE: it is still incorrect – while the market is red, it shows +20% for UM/USDC

image

### DIFF
--- a/src/pages/explore/ui/stats.tsx
+++ b/src/pages/explore/ui/stats.tsx
@@ -1,17 +1,17 @@
 'use client';
 
-// import { round } from '@penumbra-zone/types/round';
 import { Text } from '@penumbra-zone/ui/Text';
-import { InfoCard } from './info-card';
-import { pluralizeAndShortify } from '@/shared/utils/pluralize';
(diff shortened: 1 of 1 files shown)


cronokirby:
I think this is a good change to have open, but we should probably also do this in concert with revising the data on the pindexer side of things.

TalDerei:
@VanishMax can you reopen the PR in web? the corresponding issue has been stale for a while

PR #389, opened 2025-03-11
fix: #335: show volume instead of liquidity in pair selector

### LINKED ISSUE #335: Liquidity being incorrectly mapped to volume in the Pairs API
Here Liquidity is being incorrectly mapped to the volume variable.

```
      // TODO: should this be `direct_volume_over_window`?
      let volume = toValueView({
        amount: summary.liquidity,
        metadata: quoteAsset,
      });

```
https://github.com/penumbra-zone/dex-explorer/blob/main/src/shared/api/server/summary/pairs.ts#L55-L58

this manifest as the Selector widget showing liquidity but referring to it as volume. 

I'm not clear if we want to fix the API or just change volume -> liquidity everywhere.

Thoughts?

VanishMax:
Closes #335 

Uses 24 volume just like in the explore page. It is also a more performant version than what we had before

image

### DIFF
--- a/src/shared/api/server/summary/pairs.ts
+++ b/src/shared/api/server/summary/pairs.ts
@@ -9,13 +9,6 @@ import {
 import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { toValueView } from '@/shared/utils/value-view';
 import { getStablecoins } from '@/shared/utils/stables';
-import { calculateEquivalentInUSDC } from '@/shared/utils/price-conversion';
-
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
instinctively thought we should've been using [direct_volume_over_window](https://github.com/penumbra-zone/dex-explorer/pull/389/files#diff-766ee2a742a46d953112afec7e4b8d6b8055653050ff0b2b9c55593e25bfd7efL54) here, and probably should've made that change sooner – thanks for addressing 👍

Issue #387, opened 2025-03-10
Portfolio v2 - Onboarding Guide


JasonMHasperhoven:
This task should implement the [onboarding page](https://www.figma.com/design/QSQzywrHh5M6dGdVtFOY5j/%F0%9F%9F%A2-DEX-Explorer?node-id=81-120187)

<img width="730" alt="Image" src="https://github.com/user-attachments/assets/fba1c847-3878-4f74-b25f-7d6dcf931cb9" />

JasonMHasperhoven:
Moving this back into in progress for phase 1

PR #391, opened 2025-03-14
Portfolio Onboarding Section

JasonMHasperhoven:
Implements https://github.com/penumbra-zone/dex-explorer/issues/387

image

image

image


src/pages/portfolio/ui/onboarding.tsx
```diff
@@ -0,0 +1,165 @@
...
+      {footer}
+    </div>
+  );
+};
+
+export const Onboarding = observer(() => {
+  const [isOpen, setIsOpen] = useState(false);
+  const { data: providerManifests } = useProviderManifests();
+  const providerOrigins = useMemo(() => Object.keys(PenumbraClient.getProviders()), []);
+  const { isPenumbraConnected, isCosmosConnected } = useUnifiedAssets();
```
TalDerei, line 46: suggestion: can we refactor the cosmos connection status logic in `useUnifiedAssets` into a separate hook (eg. `useConnectionStatus.tsx`) to reduce scope? `useUnifiedAssets` definitionally is more focused on asset management (which includes other expensive computations like balance aggregation and price fetching).

https://github.com/penumbra-zone/dex-explorer/pull/380 can then modify `useUnifiedAssets` to use that hook accordingly. That way, we can remove all direct references to `use-unified-assets.tsx` and `use-augmented-balances.tsx` here.

src/pages/portfolio/ui/onboarding.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  const onConnectClick = () => {
```
TalDerei, line 48: comment: experiencing CORS, 401 unauthorized, and 404 not found errors in the dev logs when connecting to the keplr wallet – could this be due to it attempting to connect on localhost?

----------------

https://github.com/user-attachments/assets/2a2e5b1c-b62b-4558-bd88-c9d3934de1e0

src/pages/portfolio/hooks/use-asset-prices.tsx
```diff
@@ -0,0 +1,137 @@
...
+  price: number;
+  quoteSymbol: string;
+}
+
+/**
+ * Hook to fetch asset prices for a list of assets
+ * Uses the same summaries API that the explore page uses
+ * Only returns prices denominated in USDC
+ */
+export const useAssetPrices = (assets: Metadata[] = []) => {
```
TalDerei, line 18: ACK – this diverges from and provides a more flushed out impl than https://github.com/penumbra-zone/dex-explorer/pull/380/files#diff-2a848464ce5b56c4d4661b0e5ca75fa722410c0793592b77fd566d911b2c50d0R10. cc @vacekj for visibility

src/pages/portfolio/ui/onboarding.tsx
```diff
@@ -0,0 +1,165 @@
...
+    if (providerOrigins.length > 1) {
+      setIsOpen(true);
```
TalDerei, line 50: comment: we should avoid potentially triggering a state update during rendering – I think this violates React rules and should use `useEffect` instead?

src/pages/portfolio/hooks/use-unified-assets.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+          // Get price information
+          const priceData = prices[metadata.symbol];
+          let assetValue = 0;
+          const numericAmount = pnum(valueView).toNumber();
+
+          if (priceData?.quoteSymbol === 'USDC' && !isNaN(priceData.price)) {
+            // Use price data if available
+            assetValue = numericAmount * priceData.price;
+          } else if (isStablecoinSymbol(metadata.symbol)) {
```
TalDerei, line 118: suggestion: check if metadata is undefined

```ts
if (metadata && isStablecoinSymbol(metadata.symbol)) { ... }
```

src/features/cosmos/chain-provider.tsx
```diff
@@ -0,0 +1,79 @@
...
+  cosmwasmAminoConverters,
+  cosmwasmProtoRegistry,
+  ibcAminoConverters,
+  ibcProtoRegistry,
+  osmosisAminoConverters,
+  osmosisProtoRegistry,
+} from 'osmo-query';
+
+// @ts-expect-error type error, but it works.
+import '@interchain-ui/react/styles';
```
TalDerei, line 21: comment: unused import?

src/pages/portfolio/hooks/use-asset-prices.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+      // Fetch from summaries API
+      return apiFetch<SummaryData[]>('/api/summaries', {
+        search: '', // No search filter, get all pairs
+        limit: '200', // Increased limit to ensure we get all relevant pairs
+        offset: '0',
+        durationWindow: '1d', // Use 1-day window
+      });
+    },
+    staleTime: 60000, // 1 minute
```
TalDerei, line 39: comment: should we fetch prices more frequently?

TalDerei (commented):
great work on the onboarding section 👍

I left some observations and suggested changes to minimize the diff.

src/pages/portfolio/ui/onboarding.tsx
JasonMHasperhoven: Sounds good to me. I assume you're saying to merge first then modify? That works to reduce merge conflicts 👍

TalDerei: I was actually referring to performing this refactor first, with @vacekj resolving the minor merge conflict accordingly. Merging in the current state would _still_ result in other merge conflicts regardless.

src/pages/portfolio/ui/onboarding.tsx
TalDerei: thinking about this more, I think this may actually be a non-issue – resolving

TalDerei (approved):
optimistically approving assuming the resolution of the feedback

src/pages/portfolio/ui/onboarding.tsx
TalDerei: investigation can optionally be handled in either PR 👍

src/pages/portfolio/hooks/use-unified-assets.tsx
JasonMHasperhoven: Turns out its not needed:
image

src/pages/portfolio/ui/onboarding.tsx
JasonMHasperhoven: I think it's better to do after merging, because of merge conflicts. Noting it down.

src/pages/portfolio/ui/onboarding.tsx
JasonMHasperhoven: I think this happens indeed because of localhost issues in a prod environment. When I'm using osmosistestnet (using our testnet registry) I'm not seeing this.

src/features/cosmos/chain-provider.tsx
vacekj: Post-merge note: needed for Cosmos wallet picker to work.

PR #397, opened 2025-03-18
fix: #394: transaction pagination history

### LINKED ISSUE #394: fix pagination for transaction history
the pagination logic in https://github.com/penumbra-zone/dex-explorer/pull/382 seems to have a flaw – when the base limit is set to 20, pagination fails if no relevant transactions are detected, resulting in no transactions being displayed. However, increasing the limit to an arbitrarily higher number (eg. 50) allows transactions to start appearing iteratively. Perhaps we should differentiate between `FETCH_LIMIT` and `DISPLAY_LIMIT` instead of just using BASE_LIMIT`. cc @VanishMax 

----------------

Image

-------------

additionally, after increasing the base limit, there's a refetching flickering kind of motion at the bottom of the history page.

https://github.com/user-attachments/assets/0c3f3c8f-d7a4-4300-a937-024e188d50ca

VanishMax:
Closes #394 

Thanks to @TalDerei, the #394 issue made me reflect on the infinite scroll and generalize the `useObserver` hook.

PR contents:
- Fix to `useTransactions` hook – it was incorrectly limiting data before subaccount filtering
- Improve `useObserver` hook – it now uses throttled data fetcher, meaning that next page requests cannot fire more often than 400ms
- Remove the loader from transactions list – the request time is so fast that showing the loader only caused the fiickering

### DIFF
--- a/src/entities/position/ui/error-notice.tsx
+++ b/src/entities/position/ui/error-notice.tsx
@@ -2,7 +2,7 @@ import { BlockchainError } from '@/shared/ui/blockchain-error';
 
 export const ErrorNotice = () => {
   return (
-    <div className='min-h-screen flex items-center justify-center'>
+    <div className='min-h-[250px] flex items-center justify-center'>
       <BlockchainError
         message='An error occurred while loading data from the blockchain'
         direction='column'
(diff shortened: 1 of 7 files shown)


JasonMHasperhoven (approved):
👍

TalDerei (approved):
this fixes the issue and generally looks good to me 👍

however, I'd like to generally point out that throttling data fetching and setting a lower base limit makes loading older transactions preventively slow. eventually we might want to consider a better approach – either by streaming and displaying the entire transaction history (similar to how minifront handles it) or by segmenting transactions across individual pages.

